PR #48, opened 2012-05-07
Allow hierarchy levels to be expanded by array values

hunner:
Sometimes array values such as $roles will be used in a given hierarchy
level for Hiera. This commit will expand the level containing %{roles}
into a row for each value of $roles in order of the given array.


daenney:
This has landed in some form now: https://tickets.puppetlabs.com/browse/PUP-7204.

PR #172, opened 2014-01-08
Add output format to CLI

mrbanzai:
This allows the user to specify a return type via `-f` or `--format` (currently supports `json`, or the default `ruby`) for the Hiera results on the command-line, to allow integration with tools that don't support reading the default output format (especially when it comes to structured output).

Note that "ruby" is likely a terrible name for the default output format, but I didn't want to use "default" and couldn't think of a more appropriate format name.

Also, I didn't add any tests as I didn't see any existing tests for the CLI.

Comments / suggestions appreciated.


Zeelot:
Woo! Thanks for this! I've been using Hiera for all sorts of stuff (don't even use Puppet any more) and this makes it soooo much easier!

PR #219, opened 2014-12-04
(HI-274) Requesting deep_merge errors if absent

MikaelSmith:
Make requesting the deep or deeper behavior for the :deep_merge setting
an error if the deep_merge gem is absent.

### DIFF
--- a/lib/hiera/backend.rb
+++ b/lib/hiera/backend.rb
@@ -166,7 +166,7 @@ class Hiera
       # Merges two hashes answers with the configured merge behavior.
       #         :merge_behavior: {:native|:deep|:deeper}
       #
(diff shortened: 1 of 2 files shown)


jpartlow:
:+1: @MikaelSmith your change makes sense to me

kylog:
:+1: here too. /cc @hlindberg

hlindberg:
Hm, there was no travis run on this.. that would have caught the error that there is no Hiera.error method... wonder what happened, I was fooled by the green CLA - dammit...

MikaelSmith:
Should we implement a Hiera.error, or just revert it and say won't fix? Hiera seems to only give warnings by design...

PR #224, opened 2014-12-19
(HI-304) Make order override work with interpolation

MikaelSmith:
The existing implementation of hiera interpolation did not utilize the
order override parameter of lookup. If an interpolated variable was
expected to be looked up in the overridden datasource it would instead
only be looked up in the configured datasources.

This patch passes the order_override parameter down the call stack to
the hiera_interpolate method of the Hiera::Interpolate class. This
method then in turn passes the order_override parameter back into the
Hiera::Backend lookup method. Thereby allowing the overridden datasource
to be searched for lookup values.

### DIFF
--- a/lib/hiera/interpolate.rb
+++ b/lib/hiera/interpolate.rb
@@ -9,13 +9,15 @@ class Hiera::Interpolate
     INTERPOLATION = /%\{([^\}]*)\}/
     METHOD_INTERPOLATION = /%\{(scope|hiera|literal|alias)\(['"]([^"']*)["']\)\}/
 
-    def interpolate(data, scope, extra_data, recurse_guard, order_override)
+    def interpolate(data, scope, extra_data, context)
       if data.is_a?(String)
         # Wrapping do_interpolation in a gsub block ensures we process
         # each interpolation site in isolation using separate recursion guards.
-        recurse_guard ||= Hiera::RecursiveGuard.new
+        context ||= {}
+        new_context = context.clone
+        new_context[:recurse_guard] ||= Hiera::RecursiveGuard.new
         data.gsub(INTERPOLATION) do |match|
-          interp_val = do_interpolation(match, recurse_guard, scope, order_override, extra_data)
+          interp_val = do_interpolation(match, scope, extra_data, new_context)
 
           # Get interp method in case we are aliasing
           if data.is_a?(String) && (match = data.match(INTERPOLATION))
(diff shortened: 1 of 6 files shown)


MikaelSmith:
@hkenney or @hlindberg have comments or want to merge this? (after I rebase)

kylog:
Ping @hlindberg or @thallgren

MikaelSmith:
The test `backend.method(:lookup).arity == 4 ? Backend1xWrapper.new(backend) : backend` prevents me from simplifying the lookup interface to use an `options` hash to combine `order_override` and `recurse_guard` into a single argument.

MikaelSmith:
I've run this against Puppet specs with no failures.

MikaelSmith:
I'd be fine with context. Both recurse_guard and order_override were added to public methods, so both are options people could use. Is recurse_guard something you'd expect a hiera user to specify?

MikaelSmith:
It's meant to be a general purpose hash for parameters that may not be commonly specified, so we don't get ridiculous with the method signatures.

thallgren:
I don't expect user of hiera to specify a recurse_guard. My preference would be to create a Hiera::Backend::Context class and then pass that as a non optional argument where it's possible to do so without breaking backward compatibility and let it default to nil where needed. I'm not much in favor of hashes since they tend to be used for weird things.

thallgren:
And optional argument, well, I just did a mistake caused by that ;-)

MikaelSmith:
Missing an argument is something that's going to get worse if we keep adding optional arguments (which is the only way to do so that's backwards compatible).

thallgren:
AFAICT, the only methods that have been used by XX_backend implementations are the parse_answer and parser_string methods. These two need a ```context = nil``` parameter.

The only method that calls out is "call out" is the XX_backend.lookup method. Here we need an extra parameter but it's not optional (tested with arity, so it doesn't matter). The override can be passed twice there, once as a parameter and once in the context.

If we can agree that those three methods are the only ones that needs to be subject to backward compatibility then I suggest that the context is made non optional in the rest.

Perhaps we should consider moving 'scope' into the context as well? Duplicated in the backwards compatible methods of course but subject to change once we abandon that (there's at ticket for that).

lib/hiera/interpolate.rb
```diff
@@ -9,13 +9,14 @@ class << self
     INTERPOLATION = /%\{([^\}]*)\}/
     METHOD_INTERPOLATION = /%\{(scope|hiera|literal|alias)\(['"]([^"']*)["']\)\}/
 
-    def interpolate(data, scope, extra_data, recurse_guard)
+    def interpolate(data, scope, extra_data, options={:recurse_guard => nil, :order_override => nil})
```
MikaelSmith: This is the only area where I wasn't sure if we needed to keep backwards compatibility. It looks like we've already decided not to, so I can drop the use of a hash here and require all arguments.

MikaelSmith:
I'm ok going back to all explicit arguments for the private methods. It's slightly easier to pass opaque context objects, but you're right that it also can be used to do weird things.

thallgren:
Sounds good. I think it's either that (with the somewhat long method signatures) or doing a more elaborate effort where pass scope, override, and recurse_guard (perhaps even key when I think about it) in a context. I also prefer the former. It's clean and simple.

MikaelSmith:
I'm not sure that's worth the trouble. I'm having to tweak a lot of specs to make it happen.

MikaelSmith:
Ok, I've uploaded what I think you meant.

PR #225, opened 2015-01-09
(HI-306) Break options parsing away from Mcollective scope load

peculater:
The Hiera CLI uses ruby's optparse module to pull options out of ARGV. So
does Mcollective. The problem is that the CLI parses some options, comes
across a need for mcollective, then calls it immediately before the rest of
ARGV is parsed. That creates some weird results.  Specifically, it adds an
ordering requirement between -c and -m on the command line that is non-obvious
and undocumented.  This patch moves scope loading into a later code chunk,
after optparse is done with what it's doing.

### DIFF
--- a/bin/hiera
+++ b/bin/hiera
@@ -40,6 +40,8 @@ options = {
   :format => :ruby
 }
 
+initial_scopes = Array.new
+
 # Loads the scope from YAML or JSON files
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Thanks for your contribution, that cleans up the code nicely.

PR #227, opened 2015-01-22
(HI-327) Set acceptance tests to run on agent role

johnduarte:
This commit changes the node role that the acceptance tests run on from
`master` to `agent`.

This change is required in order for the tests to run test configurations
using the All In One agent packaging. The AIO testing applies directly to
agent nodes, and since Hiera functionality does not require a master, pinning
the test execution to the master role should be generalized to the agent.

### DIFF
--- a/acceptance_tests/tests/yaml_backend/00-setup.rb
+++ b/acceptance_tests/tests/yaml_backend/00-setup.rb
@@ -1,6 +1,7 @@
 test_name "Hiera setup for YAML backend"
 
-apply_manifest_on master, <<-PP
+agents.each do |agent|
+  apply_manifest_on agent, <<-PP
 file { '/etc/hiera.yaml':
   ensure  => present,
   content => '---
@@ -24,3 +25,4 @@ file { '/etc/puppet/hieradata':
   force   => true,
 }
 PP
+end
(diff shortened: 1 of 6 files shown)


justinstoller:
@johnduarte to put what we discussed irl here:

The file paths within the manifests that are used need to be parameterized as well because they are currently hardcoded to foss file paths. Is there a pe-hiera fork? That should have the paths either parameterized or maybe just hardcoded to the pe paths :\ 

You can see the values available currently for AIO parameterization here:
https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/unix.rb#L61-L66

Ping me when you've updated and I will re-review.

justinstoller:
Note: The content of the PR looks good against its stated scope of moving to agents. If you want we can merge this and open a separate PR for the parameterization. I know this is a part of the AIO work, why I bring the parameterization up, but I don't want to force scope creep as long as you don't mind returning to these tests later.

johnduarte:
@justinstoller 
Yes, there is a pe-facter fork, but it will die with the AIO.

The stated reason for this PR is to enable AIO testing, so it is necessary to incorporate the parameterization in it.

I have added a commit to add the parameters as necessary and have validated against a foss deployment.

kylog:
:+1: here.

PR #228, opened 2015-01-23
(maint) Don't Parameterize the hiera CLI config

justinstoller:
We parameterized most of the hard coded paths, unfortunately we don't
have a parameterization for the config hiera uses when it is ran as a
CLI tool (because it uses a different one than when it's ran as a
library, natch). This reverts the functionality of the setup step to
create the hiera.yaml file needed by the rest of the tests in the
/etc/hiera.yaml location that the CLI version of hiera expects.


justinstoller:
Whelp... My apologies for merging this, @kylog, @johnduarte. I hope this patch finds you in good spirits in the morning.

PR #230, opened 2015-01-26
(HI-328) Prevent interpolation recursion

thallgren:
This commit ensures that only one RecursiveGuard instance is
used in a lookup even if that lookup results in multiple
interpolations.

A backend implementation must add and pass on a fifth
'recurse_guard' argument but the implementation will accept
the old API with four arguments as well for backward
compatibility.

### DIFF
--- a/lib/hiera/backend.rb
+++ b/lib/hiera/backend.rb
@@ -8,6 +8,17 @@ end
 
 class Hiera
   module Backend
+    class Backend1xWrapper
+      def initialize(wrapped)
+        @wrapped = wrapped
+      end
+
+      def lookup(key, scope, order_override, resolution_type, recurse_guard)
+        Hiera.debug("Using Hiera 1.x backend API to access instance of class #{@wrapped.class.name}. Lookup recursion will not be detected")
+        @wrapped.lookup(key, scope, order_override, resolution_type)
+      end
+    end
+
     class << self
       # Data lives in /var/lib/hiera by default.  If a backend
       # supplies a datadir in the config it will be used and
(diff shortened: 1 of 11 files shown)


.gitignore
```diff
@@ -6,7 +6,7 @@ vendor/
 Gemfile.lock
 Gemfile.local
 /.project
-
 acceptance_tests/.bundle
 acceptance_tests/junit
 acceptance_tests/log
+.idea/
```
hlindberg: this is not tied to the HI.238 issue - should really be in a separate commit

lib/hiera/backend.rb
```diff
@@ -196,14 +196,20 @@ def merge_answer(left,right)
...
 
         Config[:backends].each do |backend|
           if constants.include?("#{backend.capitalize}_backend") || constants.include?("#{backend.capitalize}_backend".to_sym)
             @backends[backend] ||= Backend.const_get("#{backend.capitalize}_backend").new
-            new_answer = @backends[backend].lookup(key, scope, order_override, resolution_type)
+            backend = @backends[backend]
+            begin
+              new_answer = backend.lookup(key, scope, order_override, resolution_type, recurse_guard)
+            rescue ArgumentError
+              # Handle old style backend that take four arguments
```
hlindberg: Hm, this means it will catch argument errors that surface from an existing 5 arg lookup as well. I think it needs to look at the arity of the called lookup method instead. It also needs to issue a debug message when the 4 argument version is called. (Not sure if there is a way to issue a deprecation warning in hiera code base (that does outputs the warning once as opposed to on each call to lookup).

hlindberg: i.e. do the patching up front to avoid having to check on each call.

hlindberg:
When merged, tickets should be filed to remove the deprecation of lookup method without guard, and to remove the support for patching older backends. (I.e. target Hiera 3.0)

lib/hiera/backend.rb
```diff
@@ -8,6 +8,17 @@
 
 class Hiera
   module Backend
+    class LegacyBackendWrapper
```
hlindberg: This code is Legacy tomorrow - better to use the version. This class adapts a Hiera 1x backend to the Hiera 2x API. So... Backend1xAdapter ?

lib/hiera/backend.rb
```diff
@@ -8,6 +8,17 @@
...
+      def initialize(wrapped)
+        @wrapped = wrapped
+      end
+
+      def lookup(key, scope, order_override, resolution_type, recurse_guard)
```
hlindberg: is lookup the only method in the API?

hlindberg:
Before merging this, it is prudent to test that Puppet master and this PR works correctly together (rather than waiting until after the release of hiera 2.0.0 and upgrade of puppet to run against this version of hiera. (i.e. run puppet's spec tests against puppet and hiera source)

lib/hiera/backend.rb
thallgren: Yes. This is the only method implemented by the backends and they don't inherit, include or extend anything.

thallgren:
I have a branch of puppet/master where I have verified that puppet's spec tests run OK. No PR yet of course since it relies on adjacent hiera source and hence would fail the Travis tests.

https://github.com/thallgren/puppet/tree/issue/hi-328/fix-recursive-guard

hlindberg:
Thanks @thallgren that is great. Create a ticket for Puppet "Update puppet to hiera 2.0.0" targeting PUP 4.0.0 and add the PR there).

lib/hiera/backend.rb
hlindberg: ok - this is fine then.

PR #231, opened 2015-01-27
(HI-14) Implement structure traversal lookup

thallgren:
Previously hiera would not recognize qualified names containing
dots. This commit changes the lookup so that a key containing
dots can be used for doing qualified lookups in values that are
known to be hashes. For example:

    $ hiera user
    {"name"=>"kim", "home"=>"/home/kim"}

    $ hiera user.name
    kim

    $ hiera ssh_users
    ["root", "jeff", "gary", "hunter"]

    $ hiera ssh_users.0
    root

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,13 +85,20 @@ array.  This is used in the hiera-puppet project to replace External Node Classi
 creating a Hiera compatible include function.
 
 ### Qualified Key Lookup
-You can use a qualified key to lookup a value that is contained inside a hash:
+You can use a qualified key to lookup a value that is contained inside a hash or array:
 
 <pre>
 $ hiera user
-{"name"=>"hiera", "home"=>"/home/hiera"}
+{"name"=>"kim", "home"=>"/home/kim"}
 $ hiera user.name
-hiera
+kim
+</pre>
+
+<pre>
+$ hiera ssh_users
+["root", "jeff", "gary", "hunter"]
+$ hiera ssh_users.0
+root
 </pre>
 
 ## Future Enhancements
--- a/README.md
+++ b/README.md
@@ -84,6 +84,16 @@ Hiera can search through all the tiers in a hierarchy and merge the result into 
 array.  This is used in the hiera-puppet project to replace External Node Classifiers by
 creating a Hiera compatible include function.
 
+### Qualified Key Lookup
+You can use a qualified key to lookup a value that is contained inside a hash:
+
+<pre>
+$ hiera user
+{"name"=>"hiera", "home"=>"/home/hiera"}
+$ hiera user.name
+hiera
+</pre>
+
 ## Future Enhancements
 
  * More backends should be created
--- a/lib/hiera/backend.rb
+++ b/lib/hiera/backend.rb
@@ -197,11 +197,11 @@ class Hiera
       # databases then do so in your constructor, future calls to your
       # backend will not create new instances
 
-      # @param key [String] The key use when doing the lookup
+      # @param key [String] The key to lookup
       # @param scope [#[]] The primary source of data for substitutions.
       # @param order_override [String] An override that will be pre-pended to the hierarchy definition. Can be nil
       # @param resolution_type [Symbol] One of :hash, :array, or :priority. Can be nil which is the same as :priority
-      # return The value that corresponds to the given key or nil if no such value cannot be found
+      # @return [Object] The value that corresponds to the given key or nil if no such value cannot be found
       #
       def lookup(key, default, scope, order_override, resolution_type)
         @backends ||= {}
@@ -211,7 +211,8 @@ class Hiera
         backend_resolution_type = resolution_type
         subsegments = nil
         if segments.size > 1
-          backend_resolution_type = :hash
+          # Backend should not merge parent lookups
+          backend_resolution_type = nil
           subsegments = segments.drop(1)
         end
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -84,6 +84,16 @@ Hiera can search through all the tiers in a hierarchy and merge the result into
 array.  This is used in the hiera-puppet project to replace External Node Classifiers by
 creating a Hiera compatible include function.
 
+### Qualified Key Lookup
+You can use a qualified key to lookup a value that is contained inside a hash:
+
+<pre>
+$ hiera user
+{"name"=>"hiera", "home"=>"/home/hiera"}
```
hlindberg: using 'hiera' as the name of the user is confusing here, how about a gender neutral name like 'kim' instead ?

lib/hiera/backend.rb
```diff
@@ -196,14 +196,30 @@ def merge_answer(left,right)
       # Backend instances are cached so if you need to connect to any
       # databases then do so in your constructor, future calls to your
       # backend will not create new instances
+
+      # @param key [String] The key use when doing the lookup
```
hlindberg: Typo, "The key to lookup" perhaps?

lib/hiera/backend.rb
```diff
@@ -196,14 +196,30 @@ def merge_answer(left,right)
...
+      # @param scope [#[]] The primary source of data for substitutions.
+      # @param order_override [String] An override that will be pre-pended to the hierarchy definition. Can be nil
+      # @param resolution_type [Symbol] One of :hash, :array, or :priority. Can be nil which is the same as :priority
+      # return The value that corresponds to the given key or nil if no such value cannot be found
```
hlindberg: missing @ before return

lib/hiera/backend.rb
```diff
@@ -196,14 +196,30 @@ def merge_answer(left,right)
...
+      #
       def lookup(key, default, scope, order_override, resolution_type)
         @backends ||= {}
         answer = nil
 
+        segments = key.split('.')
+        backend_resolution_type = resolution_type
+        subsegments = nil
+        if segments.size > 1
+          backend_resolution_type = :hash
```
hlindberg: Is that really wanted here? IIRC this will do a hash merge across all layers as opposed to returning a hash bound to a key.

lib/hiera/backend.rb
```diff
@@ -196,14 +196,30 @@ def merge_answer(left,right)
...
+        end
+
         Config[:backends].each do |backend|
           if constants.include?("#{backend.capitalize}_backend") || constants.include?("#{backend.capitalize}_backend".to_sym)
             @backends[backend] ||= Backend.const_get("#{backend.capitalize}_backend").new
-            new_answer = @backends[backend].lookup(key, scope, order_override, resolution_type)
+            new_answer = @backends[backend].lookup(segments[0], scope, order_override, backend_resolution_type)
+            new_answer = qualified_lookup(subsegments, new_answer) if subsegments
 
             if not new_answer.nil?
```
hlindberg: change this 'not' to a ! while at it

lib/hiera/backend.rb
```diff
@@ -233,6 +249,16 @@ def lookup(key, default, scope, order_override, resolution_type)
       def clear!
         @backends = {}
       end
+
+      def qualified_lookup(segments, hash)
+        value = hash
+        segments.each do |segment|
+          break if value.nil? || value == :undefined
```
hlindberg: I think the use of `:undefined` is a relic stemming from some very old version of puppet. Since the new release is tied to Puppet 4.0 and it always uses nil as undefined when looking up in a scope, the check against :undefined can be dropped (unless it is an internal hiera marker of undefinedness).

If so, we never ever want a value of :undefined to be returned to puppet

lib/hiera/backend.rb
```diff
@@ -233,6 +249,16 @@ def lookup(key, default, scope, order_override, resolution_type)
...
+          raise Exception, "Hiera type mismatch: can't index #{value.class.name} with '#{segment}'" unless value.kind_of? Hash
```
hlindberg: What if the value is an array and I want the nth element ?

lib/hiera/interpolate.rb
```diff
@@ -71,11 +71,9 @@ def get_interpolation_method_and_key(data)
...
 
     def scope_interpolate(data, key, scope, extra_data)
-      value = scope[key]
-      if value.nil? || value == :undefined
-        value = extra_data[key]
-      end
-
+      segments = key.split('.')
+      value = Hiera::Backend.qualified_lookup(segments, scope)
+      value = Hiera::Backend.qualified_lookup(segments, extra_data) if value.nil? || value == :undefined
```
hlindberg: same thing for :undefined here as above

hlindberg:
Done with first review of this PR.

lib/hiera/backend.rb
thallgren: Not sure if that use case was in scope. If it is, the simplest way to implement it is probably to say that integer keys are in fact zero based indexes into an array, i.e. `x.3` will require that `x` is an array and yield the fourth element. Assumes that no strings containing integers alone are used as hiera keys. If integer hash keys are used, then we need another notation such as [] which brings on the discussion about what can be used as an argument. Alternatives are:

* Only a literal integer (hash keys use dot notation)
* An literal integer or a String, i.e. the array index `x[3]` is different from the hash key `x['3']` (now we no longer need the dot notation)
* An expression, i.e.`x[%{key.to.lookup}]`

My preference would be to keep it simple and use dot notation only, i.e. either don't permit array index or use `x.3`.

lib/hiera/backend.rb
thallgren: Although not documented there are several spec tests that adds the :undefined value. I changed so that when found, the qualified_lookup method will return nil.

hlindberg: if it is only in spec tests then that is an indication of unused functionality... will ask R.I

hlindberg: :undefined can be removed - it is from a time when it was used in Puppet. The tests for :undefined should be removed too.

lib/hiera/backend.rb
```diff
@@ -196,29 +196,46 @@ def merge_answer(left,right)
...
       def lookup(key, default, scope, order_override, resolution_type)
         @backends ||= {}
         answer = nil
 
+        segments = key.split('.')
+        backend_resolution_type = resolution_type
+        subsegments = nil
+        if segments.size > 1
+          # Backend should not merge parent lookups
+          backend_resolution_type = nil
```
hlindberg: Isn't this taking a step to far? Previous version turned :hash (merging) lookup on by default here, now it turns it off even if the user asked for it. That does not seem right - isn't it better to honor what the caller of the lookup function gave as the wanted resolution? i.e. hiera_hash function can perform a merge, produce a hash, and then segmented lookup takes place. Setting backend_resolution_type to nil denies users this ability doesn't it?

thallgren: I interpret it as the user wants to merge the hash that is found, not the parent hashes that leads up to it.

hlindberg:
We talked about this and decided that hiera_hash, and hiera_array lookups should not support segmented keys. For those, it is up to the receiver of the result to do whatever navigation that is needed as those lookup take place in a context where further processing can be made (e.g. in Ruby o Puppet). Instead, a lookup with segmented key and :hash / :array lookup should fail if the key is a segmented key.

PR #232, opened 2015-01-27
(HI-298) Change config-dir and var-dir settings

thallgren:
Changes the config_dir to '/etc/puppetlabs/agent/code' and the
var_dir to '/etc/puppetlabs/agent/code/hieradata' for posix style
systems. The corresponding paths for windows are unchanged.

The changes have been tested by running the spec tests in
current puppet/master branch while using this hiera source.

### DIFF
--- a/ext/hiera.yaml
+++ b/ext/hiera.yaml
@@ -9,7 +9,7 @@
 
 :yaml:
 # datadir is empty here, so hiera uses its defaults:
-# - /var/lib/hiera on *nix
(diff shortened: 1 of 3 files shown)


haus:
Is it worth updating https://github.com/thallgren/hiera/blob/issue/hi-298/default-dirs/ext/hiera.yaml as well?

hlindberg:
Agree with @haus - that file should be updated to reflect the changed paths

hlindberg:
The rest looks good to merge.

PR #236, opened 2015-01-28
(HI-14) Fix regression caused by incorrect assumption about Hash

thallgren:
The qualified lookup assumed that the target for the lookup was
either an Array or a Hash. That assumtpion was incorrect. It can
be either an Array or something that responds to ```[]``` (such as a
Hiera::Scope).


MikaelSmith:
:+1:

PR #238, opened 2015-01-29
(HI-298) Update hiera.yaml path in acceptance tests

johnduarte:
This commit updates the path to the hiera.yaml file in accordance with the
pathing changes in commit 602c682. This path cannot yet be parameterized and
is hard coded to `/etc/puppetlabs/agent/code/hiera.yaml` in the 00-setup.rb
test.


kurtwall:
Looks good to me, but deferring to dev for the merge.

PR #239, opened 2015-02-05
(HI-282) Fix inventory service lookups in Puppet 3

elyscape:
Puppet 3 changed the API for parsing the settings files. Puppet 3.5 changed it further. This brings those changes into Hiera for systems using Puppet 3.

### DIFF
--- a/bin/hiera
+++ b/bin/hiera
@@ -105,7 +105,8 @@ def load_scope(source, type=:yaml)
       if Puppet::Util::Package.versioncmp(puppet_version, '3.0.0') < 0
         Puppet.settings.parse
       else
-        Puppet.instance_eval { do_initialize_settings_for_run_mode(:master, []) }
+        raise "Puppet 4 does not support the inventory service" unless Puppet.respond_to?(:initialize_settings_for_run_mode)
+        Puppet.initialize_settings_for_run_mode(:master)
       end
 
       Puppet::Node::Facts.indirection.terminus_class = :rest
--- a/bin/hiera
+++ b/bin/hiera
@@ -101,11 +101,14 @@ def load_scope(source, type=:yaml)
     begin
       require 'puppet'
 
+      unless Puppet::Indirector::Terminus.terminus_classes(:facts).include?(:rest)
+        raise "rest facts terminus not present; Puppet 4+ does not support the inventory service"
+      end
+
       puppet_version = Puppet.version.split(' ').first # handle PE
       if Puppet::Util::Package.versioncmp(puppet_version, '3.0.0') < 0
         Puppet.settings.parse
       else
-        raise "Puppet 4 does not support the inventory service" unless Puppet.respond_to?(:initialize_settings_for_run_mode)
         Puppet.initialize_settings_for_run_mode(:master)
       end
 
--- a/bin/hiera
+++ b/bin/hiera
@@ -99,10 +99,21 @@ def load_scope(source, type=:yaml)
     # is fine to have the inventory_server option set even if the config
     # doesn't have the fact_terminus set to rest.
     begin
-      require 'puppet/util/run_mode'
-      $puppet_application_mode = Puppet::Util::RunMode[:master]
       require 'puppet'
-      Puppet.settings.parse
+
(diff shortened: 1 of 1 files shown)


elyscape:
A note: this would be much cleaner if either `Puppet.initialize_settings` allowed specification of the runmode or `Puppet.do_initialize_settings_for_run_mode` weren't private.

Actually, I think I have a way around that.  Let me check something.

elyscape:
There we go.  Much cleaner, though now it does rely on a private API that may change (`Puppet.do_initialize_settings_for_run_mode`).  That being said, it's remained the same since its introduction in Puppet 3.0.0.

elyscape:
Rebased onto master.  Should I squash the second commit into the first one or nuke it altogether?  It simplifies the code a lot and makes it so that we don't need to copy the contents of `Puppet.do_initialize_settings_for_run_mode` into Hiera in the event that the method changes, but it also means that we need to watch the method in case the method signature changes.  Either way involves watching it, so I'm in favor of squashing it and using the method as-is, even though it's private, but maybe you disagree.  Thoughts?

elyscape:
An alternate option would be to use `Puppet.initialize_settings_for_run_mode`, which is a public API, but it's deprecated and has been removed in the master branch.  It was deprecated in anticipation of the removal of runmodes and, IMO, its removal should have been delayed until sometime after runmodes were themselves removed.  But it has been removed on the master branch and runmodes haven't yet so we have to make do somehow.

elyscape:
Wait, I just realized something.  Since Puppet 4 removes the inventory service completely, so there's no need to have compatibility with the master branch.  If Puppet doesn't respond to that function, we can just throw an error that the inventory service isn't supported. and stop.  Problem solved!

elyscape:
As a side note, if Hiera 2 is going to be dropping support for Puppet 2.7, I can simplify this somewhat more.  Let me know and I'll make it happen.

hlindberg:
Hiera 2.0 (hiera master) only works with Puppet 4.0 and there the inventory service has been removed since it is deprecated in Puppet 3x.

melissa:
@hlindberg do you have a suggestion for moving forward? Should this PR be retargeted at the stable branch?

hlindberg:
As pointed out, the inventory system is removed in puppet 4.0 so the command needs to change. There is no maintenance release planned for hiera 1.x. As noted on the ticket, the terminus should possibly also be changed to use puppet db instead of rest. We also have problems with the hiera command, it is often overlooked when refactoring because it is not a .rb file, it needs to be restructured to call out to functionality in a .rb file (or we will continue to repeat old sins).

I understand that use case to be "test hiera structure/data against facts obtained from the "inventory". As we continue work on Puppet 4.1, we have suggested doing hiera refactoring work. I would like to hold off until then as we want to remove the circular dependency (puppet depends on a version of hiera, and hiera depends on a version of puppet). When this gets fixed I want to make sure we fix it the right way, and that the use case is well understood and supported appropriately (do we actually get all variables set by only looking up facts in the "inventory"? IIRC the real lookup gets a scope in which there could also be other values set that hiera can interpolate to form the hierarchy.  (Not to mention the additional complexity that is brought in with data in modules and environments).

I guess what I am trying to say is that we need to spend time on all of the above and we have not had the resources to do so for the hiera 2.0 release.

melissa:
Ping @hkenney, @hlindberg, @thallgren

hlindberg:
We are going to be doing work on hiera and lookup for puppet 4.1. One of the aspects of management of data is the ability to test lookups (like what I believe is the use case for this PR), but it needs to be done right. I suggest closing this PR and HI-282 in favor of a new ticket that I have yet to log.

peterhuene:
@hlindberg Did you ever end up creating a different ticket for this?  Should we close this PR?

hlindberg:
PUP 4476 is the ticket for the new command line tool "puppet lookup" which is intended to lookup key/values, handle explanations, and get facts from the node terminus etc. I think we can close this PR.

melissa:
I'm closing this ticket, as well as HI-282 in favor of PUP-4476 and associated work

PR #240, opened 2015-02-05
(HI-333) Add PuppetDB fact lookup support

elyscape:
This implements PuppetDB fact lookup support as requested in [https://tickets.puppetlabs.com/browse/HI-333](HI-333) by making use of the puppetdb facts terminus.  A direct query to PuppetDB could be done instead but would require much more code.

### DIFF
--- a/bin/hiera
+++ b/bin/hiera
@@ -111,6 +111,22 @@ def load_scope(source, type=:yaml)
         STDERR.puts "Puppet inventory service lookup failed: #{e.class}: #{e}"
         exit 1
     end
+
+  when :puppetdb
+    begin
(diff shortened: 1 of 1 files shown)


hlindberg:
This creates a circular dependency between puppet and hiera and cannot be maintained. As noted on the ticket, I also seriously question if it is a good idea to add this feature.

elyscape:
There already is a dependency on puppet in the inventory service query function, but okay. What if I implemented it as a direct query to PuppetDB?

hlindberg:
Please see the ticket - I responded there with more details what needs to be done.

PR #242, opened 2015-02-11
(maint) Revert a 1.8.7-specific security fix

kylog:


kylog:
@hlindberg @peterhuene Hiera 2.0.0 is dropping 1.8.7 support so no longer need this security fix.

PR #244, opened 2015-02-15
(HI-298) Update README to match AIO paths

joshcooper:
Updates README.md based on AIO codedir `/etc/puppetlabs/code` and vardir `/opt/puppetlabs/puppet/cache`


hlindberg:
Good catch @joshcooper - thanks for fixing.

PR #245, opened 2015-02-17
(HI-337) Distinguish between nil values and missing keys

thallgren:
This commit changes the JSON and YAML backends so that they throws
a :no_such_key exception when the key is missing and returns nil for
a key that is found but has a nil value. The Backend that calls those
backends will recognize this distinction and only use the 'extras'
hash and 'default' value when the key is missing. In contrast, a nil
return will be treated as a valid value and returned to the caller.

In addition to the hiera spec tests, this test has also been spec tested from puppet master.

### DIFF
--- a/lib/hiera/backend.rb
+++ b/lib/hiera/backend.rb
@@ -233,14 +233,20 @@ class Hiera
           subsegments = segments.drop(1)
         end
 
+        found = false
         Config[:backends].each do |backend|
           backend_constant = "#{backend.capitalize}_backend"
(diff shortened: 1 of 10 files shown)


hlindberg:
As discussed:
* Old backends cannot distinguish between producing nil, and 'not found', that must be handled in the wrapper
* Inpterpolation behavior when a nil is returned is undefined (it probably produced an empty string). After the change it is an error (not found). I think the error is preferred. There should be a spec test for that to make that behavior the expected.

PR #246, opened 2015-02-17
(#HI-338) Don't merge: This adds explain functionality

binford2k:
This adds the explain option to describe how hiera would resolve
datasources. It re-orders argument parsing slightly so this can be run
on its own, or can be output prior to the data resolution being
displayed. Context is that people find hiera lookups fiendishly
non-intuitive, so having it rendered in plain text would hopefully make
the process more consumable.

Please don't merge as-is. I'd like feedback as to whether this line of
thought would be accepted into core rather than as an external script
like we use in the classroom. If so, I'll finish it and with tests and
better error handling and a puppet DSL function and the like.

This adds this type of functionality:

```
$ RUBYLIB=./lib bin/hiera --explain -c /etc/hiera.yaml fqdn=boo
Backend data directories:
  * yaml: /etc/puppetlabs/code/hieradata

Expanded hierarchy:
  * training.puppetlabs.vm
  * boo
  * defaults

File lookup order:
  * /etc/puppetlabs/code/hieradata/training.puppetlabs.vm.yaml
  * /etc/puppetlabs/code/hieradata/boo.yaml
  * /etc/puppetlabs/code/hieradata/defaults.yaml
```

### DIFF
--- a/bin/hiera
+++ b/bin/hiera
@@ -148,6 +148,10 @@ OptionParser.new do |opts|
     options[:verbose] = true
   end
 
+  opts.on("--explain", "-e", "Describe how Hiera will be looking up data given current scope and configuration") do
+    options[:explain] = true
+  end
+
   opts.on("--array", "-a", "Return all values as an array") do
     options[:resolution_type] = :array
   end
@@ -210,7 +214,10 @@ end
 #
 # The var=val's assign scope
 unless ARGV.empty?
-  options[:key] = ARGV.delete_at(0)
+  # if this is a bareword, ie not a FACT=value argument. Allows explain without a lookup
+  unless ARGV.first =~ /^(.+?)=(.+?)$/
+    options[:key] = ARGV.delete_at(0)
+  end
 
   ARGV.each do |arg|
     if arg =~ /^(.+?)=(.+?)$/
@@ -223,9 +230,10 @@ unless ARGV.empty?
       end
     end
   end
-else
-  STDERR.puts "Please supply a data item to look up"
-  exit 1
+end
+
+unless options[:verbose]
+  Hiera.logger = "noop"
 end
 
 begin
@@ -239,10 +247,17 @@ rescue Exception => e
   end
 end
 
-unless options[:verbose]
-  Hiera.logger = "noop"
+if options[:explain]
+  hiera.explain(options[:scope])
 end
 
-ans = hiera.lookup(options[:key], options[:default], options[:scope], nil, options[:resolution_type])
-
-output_answer(ans, options[:format])
+# This conditional allow explain to run with or without a key lookup
+if options[:key]
+  ans = hiera.lookup(options[:key], options[:default], options[:scope], nil, options[:resolution_type])
+  output_answer(ans, options[:format])
+else
+  unless options[:explain]
+    STDERR.puts "Please supply a data item to look up"
+    exit 1
+  end
+end
--- a/lib/hiera.rb
+++ b/lib/hiera.rb
@@ -59,5 +59,12 @@ class Hiera
   def lookup(key, default, scope, order_override=nil, resolution_type=:priority)
     Backend.lookup(key, default, scope, order_override, resolution_type)
   end
+
+  # Just describe where Hiera would be retrieving data, given the current
+  # scope and configuration.
+  #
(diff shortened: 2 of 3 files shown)


mose:
Excellent addition. I hope it will find its way in!

kylog:
Ping @hlindberg or @thallgren

joshcooper:
ping @hlindberg @thallgren

thallgren:
I agree that this looks like a useful addition. I would like to see a more API like approach though, where the explanation is returned as a structured object (a hash). Hiera would then send the that to a method that would print the structure on stdout. Puppet however, with functions like `lookup` where hiera is only one of several contributors, could add more things to that structured object before it's presented.

hlindberg:
I would like even more of an API approach where an LookupExplainer with an API is used to report what is happening rather than just using a hash. The Explainer is passed along in calls, and those that have something to add calls methods on that object. This enables a recording of the decision making to be made and then at the end decide on what to present; i.e. "positive decisions leading to the found value" (i.e. "where did you get foo from?"), "negative decisions up to the positive found value", "did you ever consider files under the path p".

Some kind of structure approach is needed since lookups can be recursive - either by supporting recursion (pushing/popping contexts in one explainer), or by forwarding new explainers linked into a tree of explainers.

For the proposed functionality in this PR, that seems like a gross overkill, but we do have tickets for data in modules to provide just that and in a fashion that works across both global hiera data, hiera data nestd in environments and modules, as well as data provided through other means. As that is probably going to be done by the time there is a new hiera release - it seems best to focus on that (i.e. PUP-4475 add explainability).

hkenney:
Let's go ahead and close this pull request unless we're ready to work on it again. It doesn't seem actionable right now, but if we want to work on it in the future we can reopen.

law:
I would very much love a feature like this.  I've spent far too long debugging 'strange' hiera behavior only to find that someone (...hi!) forgot about an overriding parameter somewhere in a different file.  "Explain" functionality would make my life (and the lives of 'better idiots' everywhere) significantly easier.  Please reconsider adding this back into Master.

thallgren:
Puppet 4.x is equipped with a lookup command that, among several other new features, has full explanation support. It also has a "hiera" data provider built in (similar to Hiera 3.x but with full support for all lookup features).
Please read [puppet lookup Man Page](https://docs.puppetlabs.com/references/4.3.0/man/lookup.html) for more info.

binford2k:
@thallgren `puppet lookup --explain` doesn't do anything like this. All it shows is the result... basically what you'd see if you typed `hiera <thing>` on the command line. This PR is about visibility into **how** that value is resolved--which datasources are evaluated, etc.

```
root@ip-172-31-9-25:/etc/puppetlabs/code # puppet lookup message --explain
Warning: Scope(Class[Pe_console_prune]): The pe_console_prune class is deprecated. Please remove it from all of your node groups.
Data Binding "hiera"
  Found key: "message" value: "This is a message from Hiera."
```

thallgren:
Note that the Data Binding "hiera" that the explanation support tells you it's using is the very sam hiera 3.0 that we're discussing here. This data binding doesn't have any explanation support so it's no wonder that `lookup --explain` can not provide more information than that it's using it and found a key there.

If you were to use any of the new features (data in modules or environments), and the puppet "hiera" style binding there, then the explanation will be fairly elaborate. It will tell you exactly what configuration it used, how interpolated strings were compiled, and how values were merged. A sample output might look like this:
```
Merge strategy deep
  Options: {
    "merge_hash_arrays" => true
  }
  Data Binding "hiera"
    Found key: "abc::e" value: {
      "k1" => "global_g1"
    }
  Data Provider "FunctionEnvDataProvider"
    Found key: "abc::e" value: {
      "k1" => "env_e1",
      "k3" => "env_e3"
    }
  Module "abc" using Data Provider "FunctionModuleDataProvider"
    Found key: "abc::e" value: {
      "k1" => "module_e1",
      "k2" => "module_e2"
    }
  Merged result: {
    "k1" => "global_g1",
    "k2" => "module_e2",
    "k3" => "env_e3"
  }
```

binford2k:
@mose and @law I got around to porting this into a standalone gem.  I hope it helps you out. https://github.com/binford2k/hiera_explain

PR #247, opened 2015-02-18
(HI-274) Fix using non-existent Hiera#error

MikaelSmith:
Previous commit tried to use Hiera#error, which doesn't exist. Pattern
for errors is to raise an exception, so changed to raising an exception
if deep_merge is requested but unavailable. Added spec tests.

### DIFF
--- a/lib/hiera/config.rb
+++ b/lib/hiera/config.rb
@@ -54,8 +54,7 @@ class Hiera::Config
         begin
           require "deep_merge"
         rescue LoadError
-          Hiera.error "Must have 'deep_merge' gem installed for the configured merge_behavior."
-          raise
+          raise Hiera::Error, "Must have 'deep_merge' gem installed for the configured merge_behavior."
         end
       end
     end
--- a/spec/unit/config_spec.rb
+++ b/spec/unit/config_spec.rb
@@ -72,6 +72,21 @@ class Hiera
           Config.load('/dev/null')
         end
       end
+
+      describe "if deep_merge can't be loaded" do
+        let(:error_message) { "Must have 'deep_merge' gem installed for the configured merge_behavior." }
+        before(:each) do
+          Config.expects(:require).with("deep_merge").raises(LoadError, "unable to load")
+        end
+
+        it "should error if merge_behavior is 'deep'" do
+          expect { Config.load(:merge_behavior => :deep) }.to raise_error(Hiera::Error, error_message)
+        end
+
+        it "should error if merge_behavior is 'deeper'" do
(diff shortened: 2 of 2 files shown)


lib/hiera/config.rb
```diff
@@ -54,8 +54,7 @@ def validate!
         begin
           require "deep_merge"
         rescue LoadError
-          Hiera.error "Must have 'deep_merge' gem installed for the configured merge_behavior."
-          raise
+          raise "Must have 'deep_merge' gem installed for the configured merge_behavior."
```
hlindberg: this raises a RuntimeError, not a LoadError which is expected by the rspec tests. Should either raise a new LoadError, or a hiera specific error (there is one for configuration errors).

spec/unit/config_spec.rb
```diff
@@ -72,6 +72,21 @@ class Hiera
           Config.load('/dev/null')
         end
       end
+
+      describe "if deep_merge can't be loaded" do
+        let(:error_message) { "Must have 'deep_merge' gem installed for the configured merge_behavior." }
+        before(:each) do
+          Config.expects(:require).with("deep_merge").raises(LoadError, "unable to load")
```
hlindberg: see above

MikaelSmith: This causes `require "deep_merge"` to raise a LoadError, which is what's rescued above. My expectation below isn't specific about what type of error it expects, so I'll add that.

hlindberg: sorry, I misread - this mocks that the error is raised.

lib/hiera/config.rb
MikaelSmith: I'll raise the configuration error.

MikaelSmith: What is the hiera specific error for configuration errors?

MikaelSmith: InvalidConfigurationError?

hlindberg: Hiera::InvalidConfigurationError is defined if we see it as an error in the configuration. Hiera::Error if we see this as a general (standard) error.

spec/unit/config_spec.rb
```diff
@@ -72,6 +72,21 @@ class Hiera
...
+        end
+
+        it "should error if merge_behavior is 'deep'" do
+          expect { Config.load(:merge_behavior => :deep) }.to raise_error(Error, error_message)
```
hlindberg: Should be Hiera::Error

hlindberg:
Looks ok now - will merge when travis goes green.

PR #249, opened 2015-02-19
(maint) Add deep_merge gem install to git and packages pre-suite

MikaelSmith:
Puppet 4.0 requires the deep_merge gem. It needs to be available when
running the acceptance suite on git or packages against Puppet (it is
included in aio).

This adds a 'gem install deep_merge' on all hosts (for the sake of
simplicity) except windows hosts (where it is not needed).

### DIFF
--- a/acceptance_tests/setup/00_EnvSetup.rb
+++ b/acceptance_tests/setup/00_EnvSetup.rb
@@ -46,5 +46,11 @@ hosts.each do |host|
   when /solaris/
     step "#{host} Install json from rubygems"
     on host, 'gem install json'
+
+    step "#{host} Install deep_merge from rubygems"
+    on host, 'gem install deep_merge'
+  else
+    step "#{host} Install deep_merge from rubygems"
+    on host, 'gem install deep_merge'
   end
 end


kylog:
Two semi-related thoughts on this:
1) Is there something wrong in puppet if hiera needs to install puppet's dependencies?
2) Since we won't be packaging puppet 4 independently (except as a gem), does it make sense for hiera 1.x to test against puppet 4? I.e. this problem would go away if hiera/stable tested against puppet/stable.

/cc @hlindberg

MikaelSmith:
1) It seems like it. I'm not sure how testing with git expects to find gems; it seems like this is the 1st one we haven't vendored in (either in puppet-win32-ruby or puppet/lib/puppet/vendor).
2) It seems to make a lot of sense to test hiera/stable against puppet/stable for now.

kylog:
> 2) It seems to make a lot of sense to test hiera/stable against puppet/stable for now.

So my vote is we make hiera#stable green by changing it to test against puppet#stable and facter#stable for the time being (i.e. rather than this PR).

Re the deep_merge gem in puppet, I'll comment on the ticket that added that dependency.

MikaelSmith:
Put up a new PR to make Puppet and Facter branches configurable for acceptance: https://github.com/puppetlabs/hiera/pull/250

PR #250, opened 2015-02-19
(maint) Make Puppet/Facter branches configurable for acceptance

MikaelSmith:
With Puppet 4.0, we're introducing a break between master and stable
across projects. For a while, Hiera stable should run agaist
Puppet/Facter stable.

Change the acceptance Rake task to make Puppet and Facter branches
configurable.


kylog:
Works for me. I've updated Jenkins with the new variables.

If that goes green, I'll merge up to master and update that job (which will soon go away anyway).

kylog:
It was green so I merged to master (slightly awkward in git but still a 1-line change), and modified the jenkins master job. It was red already however ....

PR #251, opened 2015-02-19
(maint) Add deep_merge gem install to git and packages pre-suite

MikaelSmith:
Puppet 4.0 requires the deep_merge gem. It needs to be available when
running the acceptance suite on git or packages against Puppet (it is
included in aio).

This adds a 'gem install deep_merge' on all hosts (for the sake of
simplicity) except windows hosts (where it is not needed).


acceptance/setup/common/00_EnvSetup.rb
```diff
@@ -46,5 +46,11 @@
   when /solaris/
     step "#{host} Install json from rubygems"
     on host, 'gem install json'
+
+    step "#{host} Install deep_merge from rubygems"
+    on host, 'gem install deep_merge'
+  else
+    step "#{host} Install deep_merge from rubygems"
+    on host, 'gem install deep_merge'
```
joshcooper: I don't think we should do this in common, as that will affect `aio`. Better to restrict it to `git` and `packages`

acceptance/setup/git/pre-suite/00_EnvSetup.rb
```diff
@@ -1,50 +0,0 @@
...
-    on host, 'cd /opt/puppet-git-repos/puppet-win32-ruby; cp -r ruby/* /'
-    on host, 'cd /lib; icacls ruby /grant "Everyone:(OI)(CI)(RX)"'
-    on host, 'cd /lib; icacls ruby /reset /T'
-    on host, 'ruby --version'
-    on host, 'cmd /c gem list'
-  when /solaris/
-    step "#{host} Install json from rubygems"
-    on host, 'gem install json'
-  end
-end
```
joshcooper: I don't think we want to remove this

MikaelSmith: It's not used anywhere. Do you mean we should still have a separate git 00_EnvSetup for deep_merge?

acceptance/setup/common/00_EnvSetup.rb
MikaelSmith: Would that be true of `gem install json` as well?

kylog: I don't think we're running Solaris acceptance of hiera right now. But when we do it should be with aio packages of solaris.

joshcooper: @MikaelSmith explained things to me. He's going to update the aio/options.rb file so that this common setup step is not executed during `aio` builds. So I'm :+1:

acceptance/setup/git/pre-suite/00_EnvSetup.rb
joshcooper: @MikaelSmith explained this is a duplicate file, as there is 00_EnvSetup.rb step in common, and the one in `git` is never called. So :+1:

MikaelSmith:
Updated to remove 00_EnvSetup.rb from aio config, as there should be no environment setup for aio packages.

PR #252, opened 2015-02-20
(maint) Restore and update jenkins env

MikaelSmith:
The ci:git:test Rake task used by Jenkins looks for pre-suite
specifically in setup/git. For now we continue to maintain two setup
scripts until the Rake task is cleaned up.


MikaelSmith:
Since we vendored deep_merge, I'm going to change this to reverting the deep_merge change.

PR #255, opened 2015-02-24
(HI-298) Move hiera data to C:\ProgramData\PuppetLabs\code

joshcooper:
Previously, hiera on Windows defaulted to
C:\ProgramData\PuppetLabs\hiera\etc.

This commit updates hiera to use C:\ProgramData\PuppetLabs\code.

### DIFF
--- a/lib/hiera/util.rb
+++ b/lib/hiera/util.rb
@@ -9,7 +9,7 @@ class Hiera
 
     def microsoft_windows?
       return false unless file_alt_separator
-      
+
       begin
         require 'win32/dir'
         true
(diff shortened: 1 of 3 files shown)


joshcooper:
Wait until windows paths are finalized in https://github.com/puppetlabs/puppet-specifications/pull/25

Also, with this change, hiera won't start if it can't find it's config file. This will require changes to the MSI installer.

lib/hiera/util.rb
```diff
@@ -21,7 +21,7 @@ def microsoft_windows?
 
     def config_dir
       if microsoft_windows?
-         File.join(common_appdata, 'PuppetLabs', 'hiera', 'etc')
+         File.join(common_appdata, 'PuppetLabs', 'code')
```
ferventcoder: All hiera config is moving into code then?

joshcooper: yep

joshcooper:
This is ready to be merged

Iristyle:
It's a merge-a-thon!

PR #258, opened 2015-03-09
(HI-348) Enable deep-merge options to be passed in lookup

thallgren:
Prior to this commit it was not possible to provide deep merge options
in each call. The only way to use such options was to configure them
globally in the Hiera config. This commit changes this by allowing
the 'resolution_type' parameter to be a Hash and also ensure that is
propagated to the Backend::merge_answer() method.

When the 'resolution_type' is a hash, the merge_answer method will
completely disregard the Hiera deep merge config and instead pick the
merge behavior (keyed by :behavior) from the hash and consider the
deep merge options to the remaining top level entries of the hash.

### DIFF
--- a/lib/hiera.rb
+++ b/lib/hiera.rb
@@ -49,15 +49,67 @@ class Hiera
 
   # Calls the backends to do the actual lookup.
   #
-  # The scope can be anything that responds to [], if you have input
+  # The _scope_ can be anything that responds to `[]`, if you have input
   # data like a Puppet Scope that does not you can wrap that data in a
-  # class that has a [] method that fetches the data from your source.
+  # class that has a `[]` method that fetches the data from your source.
   # See hiera-puppet for an example of this.
   #
   # The order-override will insert as first in the hierarchy a data source
   # of your choice.
+  #
+  # Possible values for the _resolution_type_ parameter:
+  #
+  # - _:priority_ - This is the default. First found value is returned and no merge is performed
+  # - _:array_ - An array merge lookup assembles a value from every matching level of the hierarchy. It retrieves all
+  #     of the (string or array) values for a given key, then flattens them into a single array of unique values.
+  #     If _priority_ lookup can be thought of as a “default with overrides” pattern, _array_ merge lookup can be though
+  #     of as “default with additions.”
+  # - _:hash_ - A hash merge lookup assembles a value from every matching level of the hierarchy. It retrieves all of
+  #     the (hash) values for a given key, then merges the hashes into a single hash. Hash merge lookups will fail with
+  #     an error if any of the values found in the data sources are strings or arrays. It only works when every value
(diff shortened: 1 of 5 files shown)


lib/hiera/backend.rb
```diff
@@ -15,7 +15,7 @@ def initialize(wrapped)
 
       def lookup(key, scope, order_override, resolution_type, context)
         Hiera.debug("Using Hiera 1.x backend API to access instance of class #{@wrapped.class.name}. Lookup recursion will not be detected")
-        value = @wrapped.lookup(key, scope, order_override, resolution_type)
+        value = @wrapped.lookup(key, scope, order_override, resolution_type.is_a?(Hash) ? :hash : resolution_type)
```
hlindberg: wonder if this is correct - if the resolution_type is a hash isn't mapping the strategy the right thing to do? (confused...)

thallgren: The _strategy_ of the Hash (or _behavior_ in this case, since I wanted to adhere to Hiera naming conventions), is not the same thing as the _resolution_type_. The _behavior_ is either {{:native}}, {{:deep}}, or {{:deeper}} and extends how the _resolution_type_ {{:hash}} works.

hlindberg: ah ok, it is specific to this api.

hlindberg:
What happens with version 1 backends - what behavior do we get from them - i.e. is the merging behavior up to each backend - if so, then we will break them when calling with a hash. Not sure how to deal with that.

thallgren:
Version 1 backends will get :hash as the resolution_type and loose the Hash. I.e, passing a Hash will be the same as simply passing a :hash symbol. I think that's the best we can do.

lib/hiera.rb
```diff
@@ -49,15 +49,67 @@ def initialize(options={})
...
+  #          array and de-duplicated (but not automatically flattened, as in an array merge lookup).
+  #        - a hash and exists in two or more source hashes, the values from each source are recursively merged, as
+  #          though they were source hashes.
+  #        - mismatched between two or more source hashes, we haven’t validated the behavior. It should act as
+  #          described in the deep_merge gem documentation.
+  # - _deep_ - In a deep hash merge, Hiera behaves the same as for _deeper_, except that when a string/number/boolean
+  #     exists in two or more source hashes, the lowest priority value goes into the final hash. This is considered
+  #     largely useless and should be avoided. Use _deeper_ instead.
+  #
+  # The _merge_ can be given as a hash with the mandatory key `:strategy` to denote the actual strategy. This
```
nfagerlund: AFAICT from the code, it appears that the mandatory key `:strategy` does not actually exist? Is this documentation supposed to be referring to the `:behavior` key?

hlindberg: yes, I believe you found a typo in this description. It is correct in the pre-docs and implementation.

PR #264, opened 2015-04-02
(maint) Ensure consistent test locations to puppet

justinstoller:
PR #263 (61b90bfdd) didn't do what it claimed to (allow testing from PC1
repos) because the copy/pasting of Puppet/Facter code wasn't applicable
because Hiera doesn't install the test packages into the same locations
as Puppet and Facter.

This commit make Hiera use the same test package locations as Puppet and
Hiera (and thus allows the code to install PC1 packages to work).


joshcooper:
Merged in 453338f

PR #265, opened 2015-04-13
(maint) Remove Fedora 19 from build targets

melissa:
Fedora 19 went EOL on 2015-01-06. We should not longer be providing
packages for this platform.


ScottGarman:
:+1:  next reviewer, please merge soon

PR #268, opened 2015-05-14
(HI-374) configure hiera to work out of the box

mwbutcher:
Prior to this commit, users had to configure hiera.yaml
before being able to use hiera. This commit defines a datadir
and a basic hierarchy to make it easier for users to get up
and running.

please see PE-894

### DIFF
--- a/ext/hiera.yaml
+++ b/ext/hiera.yaml
@@ -2,14 +2,12 @@
 :backends:
   - yaml
 :hierarchy:
-  - defaults
-  - "%{clientcert}"
-  - "%{environment}"
-  - global
+  - "nodes/%{::trusted.certname}"
+  - common
 
 :yaml:
 # datadir is empty here, so hiera uses its defaults:
-# - /etc/puppetlabs/code/hieradata on *nix
-# - %CommonAppData%\PuppetLabs\code\hieradata on Windows
+# - /etc/puppetlabs/code/environments/%{environment}/hieradata on *nix
+# - %CommonAppData%\PuppetLabs\code\environments\%{environment}\hieradata on Windows
 # When specifying a datadir, make sure the directory exists.
   :datadir:
--- a/lib/hiera/util.rb
+++ b/lib/hiera/util.rb
@@ -29,9 +29,9 @@ class Hiera
 
     def var_dir
       if microsoft_windows?
-        File.join(common_appdata, 'PuppetLabs', 'code', 'hieradata')
+        File.join(common_appdata, 'PuppetLabs', 'code', 'environments' , '%{environment}' , 'hieradata')
       else
-        '/etc/puppetlabs/code/hieradata'
+        '/etc/puppetlabs/code/environments/%{environment}/hieradata'
       end
     end
 
--- a/spec/unit/backend_spec.rb
+++ b/spec/unit/backend_spec.rb
@@ -22,13 +22,13 @@ class Hiera
 
       it "defaults to a directory in var" do
         Config.load({})
-        Backend.datadir(:rspec, {}).should == Hiera::Util.var_dir
+        Backend.datadir(:rspec, { "environment" => "foo" }).should == Hiera::Util.var_dir % { :environment => "foo"}
 
         Config.load({:rspec => nil})
-        Backend.datadir(:rspec, {}).should == Hiera::Util.var_dir
+        Backend.datadir(:rspec, { "environment" => "foo" }).should == Hiera::Util.var_dir % { :environment => "foo"}
 
         Config.load({:rspec => {}})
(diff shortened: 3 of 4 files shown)


peterhuene:
I'm not sure if this is something that should be changed in hiera itself and not through packaging of PE / puppetserver.  At the very least it will break the default on Windows.  @melissa @kylog @hlindberg thoughts?

kylog:
I can't claim to have thought it through completely, but I tend to agree: is there a downside to having this managed as part of a PE specific module?

mwbutcher:
The idea was to leverage the aio effort and configure hiera to work 'out of
the box' for foss and PE, so using a pe module wouldn't get us there. So,
the problem would be with masterless windows hosts, right? Could we change
the hiera defaults instead?

hlindberg:
Looks like there is a chance this change will break someone that makes use of these defaults. Note that Hiera is a freestanding project and therefore anything that has to do with hiera as used with puppet has to be configured at the puppet end. (no, this is not great)

mwbutcher:
@hlindberg, that makes sense. We'll do this through the PE module instead.

ext/hiera.yaml
```diff
@@ -2,14 +2,8 @@
...
+  - "nodes/%{clientcert}"
+  - common
 
 :yaml:
-# datadir is empty here, so hiera uses its defaults:
-# - /etc/puppetlabs/code/hieradata on *nix
-# - %CommonAppData%\PuppetLabs\code\hieradata on Windows
-# When specifying a datadir, make sure the directory exists.
-  :datadir:
+  :datadir: "/etc/puppetlabs/code/environments/%{environment}/hieradata"
```
ferventcoder: Does this work with Windows?

ferventcoder: I'd state at the very least you need to keep around a comment on how to set it for Windows.

hlindberg: @ferventcoder so what should it be on windows?

ferventcoder: @hlindberg That's a great question that should probably have a followup PR to https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md

ferventcoder: Following the same logic for this, it looks like the path should be `C:\ProgramData\PuppetLabs\code\environments\%{environment}\hieradata`

Based on (https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md#puppet-agent-windows):

```md
C:\ProgramData\PuppetLabs\code *          # :codedir                     contents moved from C:\ProgramData\PuppetLabs\puppet\etc (:confdir)
    environments *                        # :environmentpath
```

ferventcoder: Something tells me this should maybe become a .yam.erb file that is translated depending on where the packaging is going.

joshcooper: And on 2003 `C:\Documents and Settings\All Users\Application Data\PuppetLabs\code`, which is why it's easier to implement this logic in code, and provide a `--configprint` like option to hiera to have it print what the defaults are.

hlindberg: Not sure if we have such a feature - it is the installer that needs to do that. @ahpook ?

ferventcoder: Thanks @joshcooper for that catch.

ext/hiera.yaml
```diff
@@ -2,14 +2,12 @@
...
-  - "%{clientcert}"
-  - "%{environment}"
-  - global
+  - "nodes/%{clientcert}"
+  - common
 
 :yaml:
 # datadir is empty here, so hiera uses its defaults:
-# - /etc/puppetlabs/code/hieradata on *nix
+# - /etc/puppetlabs/code/environments/%{environment}/hieradata on *nix
```
ferventcoder: This should be changed when the code actually gets changed to make that true. Until then it is still `/etc/puppetlabs/code/hieradata`

ferventcoder: Unless the path default fixes are coming as a separate PR.

ext/hiera.yaml
```diff
@@ -2,14 +2,12 @@
 :backends:
   - yaml
 :hierarchy:
-  - defaults
-  - "%{clientcert}"
-  - "%{environment}"
-  - global
+  - "nodes/%{clientcert}"
```
hlindberg: We should incorporate the suggested improvement from Nick Fagerlund (see ticket) and use `trusted.clientcert` instead of just `clientcert`.

nfagerlund: trusted.certname.

hlindberg:
Looks good now, merging.

PR #269, opened 2015-05-20
[DO NOT MERGE] Hiera is fragile

mwbutcher:
greatly simplifies the solution to PE-894
makes hiera more resilient by gracefully handling a missing config file

### DIFF
--- a/lib/hiera.rb
+++ b/lib/hiera.rb
@@ -40,9 +40,21 @@ class Hiera
   # If the config option is a string its assumed to be a filename,
   # else a hash of what would have been in the YAML config file
   def initialize(options={})
-    options[:config] ||= File.join(Util.config_dir, 'hiera.yaml')
+    file = File.join(Util.config_dir, 'hiera.yaml')
 
-    @config = Config.load(options[:config])
+    if File.exists?(file)
+      @config = Config.load(file)
+    else
+      @config = Config.load({
+        :backends => ["yaml"],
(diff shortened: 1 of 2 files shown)


ferventcoder:
Editing this as the commits would need to be freshened up.

lib/hiera.rb
```diff
@@ -40,9 +40,21 @@ def debug(msg); @logger.debug(msg); end
   # If the config option is a string its assumed to be a filename,
   # else a hash of what would have been in the YAML config file
   def initialize(options={})
-    options[:config] ||= File.join(Util.config_dir, 'hiera.yaml')
+    file = File.join(Util.config_dir, 'hiera.yaml')
 
-    @config = Config.load(options[:config])
+    if File.exists?(file)
```
joshcooper: This approach relies on the existence of the file or not, which means you can't create an empty `hiera.yaml` containing only comments. Ideally, it would be nice if hiera treat these cases differently:

1. the file doesn't exist, so use the defaults in code
1. it contains only comments, so use the defaults in code
1. it contains user-specified hiera configuration, so use that
1. it contains invalid yaml, and hiera should fail hard

It should be possible to detect user-specified hiera configuration, as we expect `YAML.load` to return a hash. I'm not sure about differentiating between comments-only and invalid yaml.

kylog:
Consensus has been to favor the approach in #268. So closing this one.

PR #270, opened 2015-05-24
removed defaults from hiera.yaml

crayfishx:
This is something that has irritated me for a long time now, and something I have to explain every time I run a training course or demo hiera "out of the box"...   The default hiera.yaml file has a static hierarchy entry of "default" as the first element.    This is confusing to users.  "default" implies that the value will be used if it has not been declared elsewhere, when in fact, the opposite is true - a better name would be "overrides" since anything at this level will override any other data.   Whereas, true "defaults" actually live in "global".

I think it is best completely removed as it goes against the whole concept of hierarchical overriding....

### DIFF
--- a/ext/hiera.yaml
+++ b/ext/hiera.yaml
@@ -2,7 +2,6 @@
 :backends:
   - yaml
 :hierarchy:
-  - defaults
   - "%{clientcert}"
   - "%{environment}"
   - global


peterhuene:
Hi @crayfishx.  Thanks for the contribution!  I'm not sure if this default can be removed this easily, but I do know that there is work being done to consolidate default hiera configuration and perhaps that work can drop the confusing 'defaults' entry.

@hlindberg likely knows more about this than I do.

peterhuene:
Community PR triage ping @hlindberg.

hlindberg:
HI 374 has evolved into updating the default hiera.conf. (A long discussion). I think we can close this PR in favor of the one for HI-374.

peterhuene:
Hi @crayfishx.  Thanks for the contribution!  Please see #268 which already includes this change.  As such, I'm closing this PR in favor of that one.  If you believe this is incorrect, please re-open this PR.

PR #272, opened 2015-06-10
(maint) add PE platform configs

justinstoller:
These are the configs copied from puppetlabs/puppet#4016. I've targeted master because it seems like Hiera hasn't been merged down in a while. I'm happy to repoint this at stable if I should.

===begin original commit msg===
Prior to this there were no acceptance test host config files for PE
only platforms. The distinction between PE and FOSS is largely going
away (save for where the failures are reported).

This commit adds the necessary host configs to test Hiera against all
the platforms we will distribute puppet-agent on in the Shallow Gravy
timeframe.

### DIFF
--- a/acceptance/config/nodes/centos-4-i386.yaml
+++ b/acceptance/config/nodes/centos-4-i386.yaml
@@ -1,0 +1,19 @@
+---
+HOSTS:
+  master:
+    roles:
+    - master
+    platform: el-7-x86_64
+    hypervisor: vcloud
+    template: redhat-7-x86_64
+  agent:
+    roles:
+    - agent
+    platform: centos-4-i386
+    hypervisor: vcloud
+    template: centos-4-i386
+CONFIG:
+  datastore: instance0
+  resourcepool: delivery/Quality Assurance/FOSS/Dynamic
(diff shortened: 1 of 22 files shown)


joshcooper:
ping @justinstoller @hlindberg is this needed? should we target stable?

hlindberg:
the PR was from 20 days ago, at that point it was correct to target master. Master was merged down to stable as hiera 3.0.0 was released - so yes, this should target stable (and then be merged to master).

MikaelSmith:
How much does this overlap with https://github.com/puppetlabs/hiera/pull/277, and do we really need the masters setup?

kevpl:
in my runs in #277, I was able to get success without masters, so I would suggest not using them if possible (I'm not very familiar w/hiera testing at this point).

Seeing that this is just node definitions, I would suggest pushing through my PR, closing this one, and I can remake this into just-agent host files for these platforms.  What do you guys think?

MikaelSmith:
Sounds like a plan to me, I'll do a little verification of #277 and merge it.

MikaelSmith:
Closing in favor of @kevpl finishing off the configs with agent-only files.

kevpl:
for tracking, that finishing work is in #278

PR #273, opened 2015-06-15
(HI-369) Remove mock and cow build targets

melissa:
With the switch over to delivering software via the puppet agent
packages, we no longer need to maintain this list of build targets for
use by the packaging repo automation. We do want to keep some of this
around and available in the off chance that a community member has the
desire to build their own packages using the packaging repo automation.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -1,23 +1,21 @@
 ---
 packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
 packaging_repo: 'packaging'
-default_cow: 'base-squeeze-i386.cow'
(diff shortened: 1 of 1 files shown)


haus:
@melissa not stable?

melissa:
@haus master is where the 3.0.0 tag is. Stable is still the 2.x series, as far as I'm aware

haus:
@melissa okay, but we haven't been shipping packages for 2.x of hiera either

melissa:
hmm, I forgot about that. Alright, I can retarget this at stable in that case. Just for clarification though, why is it that we didn't do the full ship of packages for hiera 2.x again?

haus:
because it was part of puppet-agent from the getgo.

melissa:
Closed in favor of https://github.com/puppetlabs/hiera/pull/274

PR #274, opened 2015-06-16
(HI-369) Remove mock and cow build targets

melissa:
With the switch over to delivering software via the puppet agent
packages, we no longer need to maintain this list of build targets for
use by the packaging repo automation. We do want to keep some of this
around and available in the off chance that a community member has the
desire to build their own packages using the packaging repo automation.


haus:
:+1:

ScottGarman:
Putting the COWs out to pasture, I see.

shrug:
@ScottGarman Don't MOCK @melissa

PR #275, opened 2015-06-22
(HI-397) Add missing nodes/ entry to in code config defaults

hlindberg:
Before this the defaults expressed in the ext/hiera.yaml and the
internal code were different.

This makes them have the same content.

### DIFF
--- a/lib/hiera/config.rb
+++ b/lib/hiera/config.rb
@@ -4,14 +4,14 @@ class Hiera::Config
     # load takes a string or hash as input, strings are treated as filenames
     # hashes are stored as data that would have been in the config file
     #
-    # Unless specified it will only use YAML as backend with a single
-    # 'common' hierarchy and console logger
+    # Unless specified it will only use YAML as backend with a
+    # hierarchy of 'nodes/%{::trusted.certname}' and 'common', and with a
+    # console logger.
     #
-    # @return [Hash] representing the configuration.  e.g.
-    #   {:backends => "yaml", :hierarchy => "common"}
+    # @return [Hash] representing the configuration.
     def load(source)
-      @config = {:backends => "yaml",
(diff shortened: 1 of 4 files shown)


lib/hiera/config.rb
```diff
@@ -10,8 +10,8 @@ class << self
     # @return [Hash] representing the configuration.  e.g.
     #   {:backends => "yaml", :hierarchy => "common"}
```
joshcooper: could you update this comment, and a comment above this about what the default behavior is supposed to be?

lib/hiera/config.rb
```diff
@@ -10,8 +10,8 @@ class << self
...
     def load(source)
-      @config = {:backends => "yaml",
-                 :hierarchy => "common",
+      @config = {:backends => ["yaml"],
+                 :hierarchy => ["nodes/%{::trusted.certname}", "common"],
```
joshcooper: does Hiera have any tests for this sort of thing, e.g. what is the state of `config` after `load` is called?

joshcooper: Looks like it does https://travis-ci.org/puppetlabs/hiera/jobs/67921209#L428-L429

hlindberg: apparently :-) fixing

lib/hiera/config.rb
hlindberg: fixing

hlindberg:
@joshcooper comments addressed and tests updated

joshcooper:
I left comments in the ticket, but the short version is I don't think this is a blocker for the release, as it only happens when running puppet as non-root and there is no per-user default hiera.yaml in `~/.puppetlabs/etc/code/hiera.yaml`. It also proves the point that defaults should be expressed in code as we originally talked about.

PR #277, opened 2015-07-10
(PUP-4757) added :peaio task to test puppet-agent on PE-only platforms

kevpl:


acceptance/Rakefile
```diff
@@ -78,6 +78,11 @@ def beaker_test(mode = :packages, options = {})
     final_options[:install] << "#{build_giturl('hiera')}##{sha}"
   end
 
+  if mode == :peaio
+    final_options[:puppet_agent_sha] = sha
+    final_options[:puppet_agent_version] = sha
+  end
+
```
joshcooper: Does this mean you can only test tags, e.g. `2015.2.0`, and not individual commits (since I think `puppet_agent_sha` needs to be the full sha and `puppet_agent_version` needs to be the git-describe)? I also don't know how the PE pipeline works...

MikaelSmith: It maps to agent-download versions (agent-downloads.delivery...), which seem to have their own convention - puppet-agent releases with an additional tag, like `1.2.1.41`.

MikaelSmith:
This worked in my testing with `SHA=1.2.1.41` using SLES12.

PR #279, opened 2015-07-17
(PUP-4757) updated :peaio build location

kevpl:
before, :peaio was set to build from the pe_promoted repo.
this was a misunderstanding of the context of the pipeline
that this task is used in. this commit fixes that so that
it gets puppet-agent from the builds server.

[skip ci]


kevpl:
I've tested this against the Jenkins hiera acceptance job successfully [here](https://jenkins.puppetlabs.com/view/PE%20AIO/job/platform_aio-suite_intn-sys_stable-pe-hiera/29/) with this change.

PR #283, opened 2015-07-30
(HI-184) Inconsistent handling of fully-qualified facts when called from the command line [#puppethack]

rnelson0:
Add a default_prof to the scope hash to create a matching no-colon-prepend key when a prepended key exists.

### DIFF
--- a/acceptance/tests/test-184-command-line-facts.rb
+++ b/acceptance/tests/test-184-command-line-facts.rb
@@ -1,0 +1,66 @@
+test_name "Command-line lookup should resolve based on fact values from facter run"
+
+agents.each do |agent|
+  codedir = agent.puppet['codedir']
+  hieradatadir = File.join(codedir, 'hieradata')
+
+  teardown do
+    apply_manifest_on agent, <<-PP
+    file { '#{hieradatadir}':
+      ensure  => directory,
+      recurse => true,
+      purge   => true,
+      force   => true,
+    }
+    file { '#{codedir}/scope.yaml': ensure => absent }
+    file { '#{codedir}/scope.json': ensure => absent }
+    PP
+  end
+
+  osfamily = (on agent, facter('osfamily')).stdout.chomp
+
+  apply_manifest_on agent, <<-PP
+file { '#{codedir}/hiera.yaml':
+  ensure  => present,
+  content => '---
+    :backends:
+      - "yaml"
+    :logger: "console"
+    :hierarchy:
+      - "osfamily/%{::osfamily}"
+      - "%{environment}"
+      - "global"
+
+    :yaml:
+      :datadir: "#{hieradatadir}"
+  '
+}
+
+file { ['#{hieradatadir}', '#{hieradatadir}/osfamily']:
+  ensure => directory,
+}
+
+file { '#{hieradatadir}/global.yaml':
+  ensure => present,
+  content => "---
+    foo: zan
+  "
+}
+
+file { '#{hieradatadir}/osfamily/#{osfamily}.yaml':
+  ensure => present,
+  content => "---
+    foo: bar
+  "
+}
+PP
+
+  on agent, hiera('foo', "osfamily=#{osfamily}") do
+    assert_match /bar/, result.stdout
+  end
+
+  on agent, hiera('foo', "::osfamily=#{osfamily}") do
+    assert_match /bar/, result.stdout
+  end
+
+end
--- a/acceptance/ubuntu.yaml
+++ b/acceptance/ubuntu.yaml
@@ -1,0 +1,10 @@
+# Docker
+HOSTS:
+  ubuntu:
+    platform: ubuntu-14.04-x64
+    image: ubuntu
+    hypervisor: docker
+    roles:
+    - agent
+CONFIG:
+  type: foss
--- a/lib/hiera/backend.rb
+++ b/lib/hiera/backend.rb
@@ -307,8 +307,19 @@ class Hiera
 
       def qualified_lookup(segments, hash)
         value = hash
+
         segments.each do |segment|
           throw :no_such_key if value.nil?
+          if segment =~ /^::(.+)/
+            othersegment = $1
+          else
+            othersegment = "::%s" % segment
+          end
+
+          if value.include?(othersegment)
+            value[segment] = value[othersegment]
+          end
+
           if segment =~ /^[0-9]+$/
             segment = segment.to_i
             raise Exception, "Hiera type mismatch: Got #{value.class.name} when Array was expected enable lookup using key '#{segment}'" unless value.instance_of?(Array)
--- a/spec/unit/backend_spec.rb
+++ b/spec/unit/backend_spec.rb
@@ -212,12 +212,12 @@ class Hiera
       end
 
       @leading_double_colon_tests = {
-        "test_%{::rspec::data}_test" => "test__test",
-        "test_%{scope('::rspec::data')}_test" => "test__test"
+        "test_%{::rspec::data}_test" => "test_value_test",
+        "test_%{scope('::rspec::data')}_test" => "test_value_test"
       }
 
       @leading_double_colon_tests.each do |input, expected|
-        it "does not try removing leading :: when a full lookup fails (#17434)" do
+        it "does find a matching key with leading :: when a full lookup fails (#17434)" do
           expect(Backend.parse_string(input, {"rspec::data" => "value"})).to eq(expected)
         end
       end
--- a/tmp/hiera/common.yaml
+++ b/tmp/hiera/common.yaml
@@ -1,0 +1,2 @@
+---
+key: Commonvalue
--- a/tmp/hiera/hiera.yaml
+++ b/tmp/hiera/hiera.yaml
@@ -1,0 +1,6 @@
+:hierarchy:
+  - "osfamily_%{::osfamily}"
+  - "%{::fqdn}"
+  - common
+:yaml:
+   :datadir: /var/hiera
--- a/tmp/hiera/host.example.com.yaml
+++ b/tmp/hiera/host.example.com.yaml
@@ -1,0 +1,2 @@
+---
+key: Hostvalue
--- a/tmp/hiera/osfamily_RedHat.yaml
+++ b/tmp/hiera/osfamily_RedHat.yaml
@@ -1,0 +1,2 @@
+---
+key: RedHatvalue
--- a/tmp/hiera/other.yaml
+++ b/tmp/hiera/other.yaml
@@ -1,0 +1,2 @@
+---
+key: Othervalue
--- a/tmp/setup
+++ b/tmp/setup
@@ -1,0 +1,4 @@
+#!/bin/sh
+sudo mkdir /var/hiera
+sudo chown $USER /var/hiera
+cp tmp/hiera/* /var/hiera
--- a/tmp/test
+++ b/tmp/test
@@ -1,0 +1,47 @@
+#!/bin/sh
+HIERA='bundle exec bin/hiera -c /var/hiera/hiera.yaml'
+
+echo ""
+echo "Basic test, no scope"
+echo -n "expect Commonvalue: got "
+$HIERA key
+
+echo ""
+echo "osfamily=Debian"
+echo -n "expect Commonvalue: got "
+$HIERA key osfamily=Debian
+
+echo ""
+echo "osfamily=RedHat"
+echo -n "expect RedHatvalue: got "
+$HIERA key osfamily=RedHat
+
+echo ""
+echo "fqdn=other"
+echo -n "expect Othervalue: got "
+$HIERA key fqdn=other
+
+echo ""
+echo "fqdn=host.example.com"
+echo -n "expect Commonvalue: got "
+$HIERA key fqdn=host.example.com
+
+echo ""
+echo "::osfamily=Debian"
+echo -n "expect Commonvalue: got "
+$HIERA key ::osfamily=Debian
+
+echo ""
+echo "::osfamily=RedHat"
+echo -n "expect RedHatvalue: got "
+$HIERA key ::osfamily=RedHat
+
+echo ""
+echo "::fqdn=other"
+echo -n "expect Othervalue: got "
+$HIERA key ::fqdn=other
+
+echo ""
+echo "::fqdn=host.example.com"
+echo -n "expect Commonvalue: got "
+$HIERA key ::fqdn=host.example.com


rnelson0:
Thanks to @ripienaar for the assistance on this.

This change addresses HI-184. I found that rake -T did not show a spec target, though rake spec did work, so I am unsure what I would need to do to properly add a test to cover this. I'll look into it a bit and add to the PR if I find something. If there's a guide for this, or a clue why rake -T doesn't work, please let me know.

MikaelSmith:
Tried writing an acceptance test (https://github.com/puppetlabs/puppet/blob/master/docs/acceptance_tests.md). For some reason I can't reproduce the fixed behavior. The test is

```
test_name "Command-line lookup should resolve based on fact values from facter run"

agents.each do |agent|
  codedir = agent.puppet['codedir']
  hieradatadir = File.join(codedir, 'hieradata')

  teardown do
    apply_manifest_on agent, <<-PP
    file { '#{hieradatadir}':
      ensure  => directory,
      recurse => true,
      purge   => true,
      force   => true,
    }
    file { '#{codedir}/scope.yaml': ensure => absent }
    file { '#{codedir}/scope.json': ensure => absent }
    PP
  end

  osfamily = (on agent, facter('osfamily')).stdout.chomp

  apply_manifest_on agent, <<-PP
file { '#{codedir}/hiera.yaml':
  ensure  => present,
  content => '---
    :backends:
      - "yaml"
    :logger: "console"
    :hierarchy:
      - "osfamily/%{::osfamily}"
      - "%{environment}"
      - "global"

    :yaml:
      :datadir: "#{hieradatadir}"
  '
}

file { ['#{hieradatadir}', '#{hieradatadir}/osfamily']:
  ensure => directory,
}

file { '#{hieradatadir}/global.yaml':
  ensure => present,
  content => "---
    foo: zan
  "
}

file { '#{hieradatadir}/osfamily/#{osfamily}.yaml':
  ensure => present,
  content => "---
    foo: bar
  "
}
PP

  on agent, hiera('foo', "osfamily=#{osfamily}") do
    assert_match /bar/, result.stdout
  end

  on agent, hiera('foo', "::osfamily=#{osfamily}") do
    assert_match /bar/, result.stdout
  end

end
```

In my environment, I run with `be beaker --options-file merged_options.rb --hosts=config/nodes/debian-7-x86_64.yaml --preserve-hosts --tests=tests/test-184-command-line-facts.rb`.

rnelson0:
Also ran the ruby suggested at https://gist.github.com/ripienaar/62b199d83e87fa0051d4 on my machine and in the beaker test nodes with no issues.

rnelson0:
@MikaelSmith Here's a [gist](https://gist.github.com/rnelson0/a9de7d31c8983a8536cd) with the hiera.yaml, individual .yaml files, the tree layout (since gists won't let you path properly), and the output I receive. This is on CentOS release 6.4 (Final) using RVM with ruby 1.9.3p551 (2014-11-13 revision 48407) [x86_64-linux].

HAIL9000:
Hey @rnelson0 have you gotten a chance to consult with @ripienaar about this acceptance test yet?

ripienaar:
I've unfortunately not had time to take a proper look

MikaelSmith:
I've setup a CentOS 6.6 docker machine with Ruby 1.9.3p551 using RVM, and this patch, and I still can't reproduce the fixed behavior @rnelson0 is seeing.

ripienaar:
OK, it's failing because https://github.com/puppetlabs/hiera/blob/master/lib/hiera/backend.rb#L311 checks if the particular key is ```include?```'d in the hash which our tweak wont address.

Why we could get it working here I dont have time to look at now but suffice to say we need to do some more fixing before this can be merged

whopper:
Hey @rnelson0! Just checking in to see if you'll have any time to continue on this.

rnelson0:
@whopper I would love to but don't think I'll have time until after PuppetConf - or maybe at the contributor's summit there. Lots to do in ~3 weeks to prepare!

whopper:
@rnelson0 no worries! We'll probably see you at the contributor's summit too! :)

whopper:
Hey @rnelson0! Just checking in to see if you're still interested in working on this!

Hope you had a great Holiday!

rnelson0:
Wow, time flies! I'm game for working on this if someone can help point me in the right direction for the beaker tests, which appears to highlight the issue. Is that something we can figure out on IRC?

whopper:
@rnelson0 actually we are having our PR triage tomorrow at 10:00AM PDX time, if you happen to be available to hang out and talk about the beaker issues for a little bit. If not, I'll put out an internal ping to see who'd have the most context to help you figure out what's going on, and when we'd have time to bring it in for work.

rnelson0:
Unfortunately, tomorrow is no good for me. I'll be around either the 12th
or 19th if we can't find some other time to work it in before then.

whopper:
Mk - well I moved this PR to the 'discussion' section for tomorrow, so we'll talk about it and do proper triaging for where it'll land. Stay tuned!

MikaelSmith:
The acceptance test is just in the comment I put above. To run tests using Docker or Vagrant (using Docker with Beaker on Mac OS X is a pain)
- Change to `hiera/acceptance`
- Copy the test above into `./tests/test-184-command-line-facts.rb`
- Install gems: `bundle install --path=.bundle/gems`
- Create an centos7.yaml file containing:
```
# Docker
HOSTS:
  centos-7-64:
    platform: centos-7-x64
    image: centos:7
    hypervisor: docker
    roles:
    - agent
CONFIG:
  type: foss
```
or
```
# Vagrant
HOSTS:
  centos-7-64:
    platform: centos-7-x64
    box: puppetlabs/centos-7.2-64-nocm
    hypervisor: vagrant
    roles:
    - agent
CONFIG:
  type: foss
```
Then when neither of those work, start the Vagrant machine manually and use
```
HOSTS:
  centos-7-64:
    platform: centos-7-x64
    roles:
    - agent
CONFIG:
  type: foss
```
- Run tests with `bundle exec beaker --options config/git/options.rb --hosts centos7.yaml --tests tests/test-184-command-line-facts.rb --load-path lib --install "PUPPET/master,FACTER/2.x,https://github.com/rnelson0/hiera.git#normalize_cli_keys"`. You may need to specify `--keyfile ~/.ssh/a_key_file` as well, I'm not sure what the default behavior is.

You should see a failure along the lines of
```
Failed Tests Cases:
  Test Case tests/test-184-command-line-facts.rb reported: #<Minitest::Assertion: Expected /bar/ to match "zan\n".>
```

rnelson0:
@MikaelSmith @ripienaar Very interesting. First, I added `tmp/setup` and `tmp/test` that can be used to set up some hiera files in `/var/hiera/` and then used for a test. You can check out that commit, I inserted it prior to the other changes, and then move forward commit by commit to see what happens.

On the PR triage today, we verified L311 looked like the culprit, so I inserted the `default_proc` at [L302](https://github.com/rnelson0/hiera/blob/normalize_cli_keys/lib/hiera/backend.rb#L302) and was not seeing any changes:
````
$ tmp/test

Basic test, no scope
expect Commonvalue: got Commonvalue

osfamily=Debian
expect Commonvalue: got Commonvalue

osfamily=RedHat
expect RedHatvalue: got Commonvalue

fqdn=other
expect Othervalue: got Commonvalue

fqdn=host.example.com
expect Commonvalue: got Commonvalue

::osfamily=Debian
expect Commonvalue: got Commonvalue

::osfamily=RedHat
expect RedHatvalue: got RedHatvalue

::fqdn=other
expect Othervalue: got Othervalue

::fqdn=host.example.com
expect Commonvalue: got Hostvalue
````
So I added some debugging around the original [L303](https://github.com/rnelson0/hiera/blob/normalize_cli_keys/lib/hiera/backend.rb#L303), resulting in:
````
      def qualified_lookup(segments, hash)
        value = hash
        value.default_proc = ->(hash, key) {        #L303
          if key =~ /^::(.+)/
            otherkey =$1
          else
            otherkey = "::#{key}"
          end

          if hash.include?(otherkey)
            hash[key] = hash[otherkey]
          else
            hash[otherkey] = hash[key]        #L313
          end
        }

        segments.each do |segment|
          if segment =~ /osfam/
            puts ""
            puts "value of segment: #{segment}"
            puts "value of osfamily: #{value['osfamily']}"
            puts "value of ::osfamily: #{value['::osfamily']}"
          end
          throw :no_such_key if value.nil?
          ...
````
Here's where it gets interesting:
````
$ tmp/test

Basic test, no scope
expect Commonvalue: got
value of segment: ::osfamily
/home/rnelson0/git/hiera/lib/hiera/backend.rb:303: stack level too deep (SystemStackError)

osfamily=Debian
expect Commonvalue: got
value of segment: ::osfamily
value of osfamily: Debian
value of ::osfamily: Debian
Commonvalue

osfamily=RedHat
expect RedHatvalue: got
value of segment: ::osfamily
value of osfamily: RedHat
value of ::osfamily: RedHat
RedHatvalue

fqdn=other
expect Othervalue: got
value of segment: ::osfamily
/home/rnelson0/git/hiera/lib/hiera/backend.rb:313: stack level too deep (SystemStackError)

fqdn=host.example.com
expect Commonvalue: got
value of segment: ::osfamily
/home/rnelson0/git/hiera/lib/hiera/backend.rb:304: stack level too deep (SystemStackError)

::osfamily=Debian
expect Commonvalue: got
value of segment: ::osfamily
value of osfamily: Debian
value of ::osfamily: Debian
Commonvalue

::osfamily=RedHat
expect RedHatvalue: got
value of segment: ::osfamily
value of osfamily: RedHat
value of ::osfamily: RedHat
RedHatvalue

::fqdn=other
expect Othervalue: got
value of segment: ::osfamily
/home/rnelson0/git/hiera/lib/hiera/backend.rb:303: stack level too deep (SystemStackError)

::fqdn=host.example.com
expect Commonvalue: got
value of segment: ::osfamily
/home/rnelson0/git/hiera/lib/hiera/backend.rb:313: stack level too deep (SystemStackError)
````
Lines 303/313 are marked above. I'm not sure why, but adding debugging appears to introduce a heisenbug. If you run this a few times, you'll see that the errors are random. Eventually I got `tmp/test: line 42:  2984 Segmentation fault      $HIERA key ::fqdn=other` on some lookups. Sometimes it totally goes out to lunch and I have to CTRL-Z; kill %1 to end it.

**Comment by rnelson0** [comment]
OK, some more poking at it makes me think I got it. Rather than changing `default_proc` I simply made sure the "other" key is available. The docker test passes now. I'm not sure the original fix is required, then.

**Comment by MikaelSmith** [comment]
Probably don't want to add all the files in `tmp/`.

File: lib/hiera/backend.rb [hunk 1]
```diff
@@ -300,8 +300,19 @@ def clear!
...
+	    othersegment = $1
+	  else
+	    othersegment = "::%s" % segment
+	  end
+
+	  if value.include?(othersegment)
+	    value[segment] = value[othersegment]
+	  end
+
+	  throw :no_such_key if value.nil?
```
MikaelSmith: You want this before value.include?, or you'll cause an error to be thrown about value being nil.

rnelson0:
@MikaelSmith I'll remove that when I rebase it.

rnelson0:
@MikaelSmith I think commit 52d63cc is probably not needed, and in fact doesn't seem to affect anything. Of course the two temporary commits can go and then the remaining three can be squashed down to one. Before I whack that commit, do you see any reason to keep it around?

bin/hiera
```diff
@@ -243,6 +243,19 @@ unless options[:verbose]
   Hiera.logger = "noop"
 end
 
+# Allow lookup of keys with or without prepended double colons
```
MikaelSmith: This doesn't seem necessary to fix the issue we've seen. I don't understand how the scope is used sufficiently to know whether there's a case we haven't thought of, but I lean towards doing the minimal fix until we come up with a case that needs this.

rnelson0:
@hlindberg We discussed this on the Core Triage today and this is ready for your review now! Let me know if you have questions. Thanks!

hlindberg:
I have problems with the change. It is the command line tool that seems to be at fault and not the hiera backend which I think should remain unchanged. I based this on the problem description in the ticket.

The tests are also failing.

I am also wondering if any of this is worth the time/trouble in the first place. (I.e. making changes to the hiera command line tool) since the new `puppet lookup` command works, is what will be used in the future, and has more options than the old hiera commad line tool (which is broken in several other ways as well).

spec/unit/backend_spec.rb
```diff
@@ -212,12 +212,12 @@ def lookup(key, scope, order_override, resolution_type)
       end
 
       @leading_double_colon_tests = {
-        "test_%{::rspec::data}_test" => "test__test",
-        "test_%{scope('::rspec::data')}_test" => "test__test"
+        "test_%{::rspec::data}_test" => "test_value_test",
```
MikaelSmith: I agree with Henrik, changing this test looks strange.

lib/hiera/backend.rb
```diff
@@ -307,8 +307,19 @@ def clear!
 
       def qualified_lookup(segments, hash)
         value = hash
+
         segments.each do |segment|
           throw :no_such_key if value.nil?
+          if segment =~ /^::(.+)/
```
MikaelSmith: Now that I think about it, we should be able to fix this from `bin/hiera`. Going to try just adding entries to options[:scope] rather than overriding default_proc.

MikaelSmith: Putting
```
options[:scope].keys.each do |key|
  if key =~ /^::(.+)/
    otherkey = $1
  else
    otherkey = "::%s" % key
  end
  options[:scope][otherkey] = options[:scope][key]
end
```
in bin/hiera works. `default_proc` doesn't appear to be called on `include?`.

acceptance/ubuntu.yaml
```diff
@@ -0,0 +1,10 @@
+# Docker
```
MikaelSmith: Remove this from the final commit.

acceptance/tests/test-184-command-line-facts.rb
```diff
@@ -0,0 +1,66 @@
...
+
+file { '#{hieradatadir}/osfamily/#{osfamily}.yaml':
+  ensure => present,
+  content => "---
+    foo: bar
+  "
+}
+PP
+
+  on agent, hiera('foo', "osfamily=#{osfamily}") do
```
MikaelSmith: It would be nice to extend the acceptance test with all the test cases from the `tmp` dir. Then remove the `tmp` dir.

hlindberg:
As I said earlier and on the ticket - is it really worth spending time on this? A usability feature on a semi broken command? Remember that It will be bundled with puppet versions where the lookup command exists, and it works with hiera data, environment data and data in modules, the hiera command line tool only with global hiera data. I rather fix issues in lookup if there is something lookup cannot do. We have invested a lot of development time and QA time to ensure it works well - the hiera command line tool is in stark contrast; there is always something wrong with it.

rnelson0:
If this fix can only be ported to hiera 3, then it appears to be a moot point. If it can be ported to hiera 1.x, then admins still using puppet 3 may rejoice.

hlindberg:
Even if i is possible to backport a fix, we are not planning to release any new versions except if there are security issues. We will do a 3.0.7 shortly. Release of 3.1.0 is not yet determined, and this PR currently targets master which is for  3.1.0.

rnelson0:
@hlindberg With that in mind, if you think `puppet lookup` is the best place to focus this, it sounds like the PR should be closed and that you're inclined to close the HI-184 ticket, too. I am okay with that, I learned a lot anyway. I'll have to spend some time with `puppet lookup` soon and make sure things work right there!

hlindberg:
@rnelson0 yes, I think the energy is best spent on the new lookup implementation (command line tool, and function) for which we have a substantial testing suite, proper documentation and APIs. It may not be perfect yet as it is not tested i battle, and it may not support all the wanted use cases. But we want to hear about those and add them.

joshcooper:
@rnelson0 Thanks for working on this PR. It sounds like you and @hlindberg have a path forward using `puppet lookup` so I'm going to close this PR and ticket. It would be great to get your feedback with `lookup` and make sure it satisfies the use cases you care about!

rnelson0:
@joshcooper will do!

PR #285, opened 2015-07-31
WIP (maints) Fix commits checker regex

MikaelSmith:
Previously the check for the trailing paren was only made
for HI-<digits> (but not for maint or doc or packaging). This fixes
that regex.


MikaelSmith:
Verifying it works.

PR #289, opened 2015-08-18
(maint) remove overloaded git functions from acceptance lib

er0ck:
This change removes build_giturl() and lookup_in_env() from the puppet
acceptance library. These methods have been moved to beaker where they
can be properly tested and maintained.
This change also uses the new method names, which have been updated to
conform to newer beaker standards.

[skip ci]


MikaelSmith:
Have you tried running with ci:test:git? If I use RedHat 6, it fails because Puppet 4 requires Ruby 1.9.3+

MikaelSmith:
Works with RedHat 7 of course.

PR #294, opened 2015-09-03
(HI-428) Add Solaris 10/11 SPARC node defs

johnduarte:
This commit adds node definitions for solaris sparc 10 and 11
agents using the SPARC architecture.

### DIFF
--- a/acceptance/config/nodes/solaris-10-sparc.yaml
+++ b/acceptance/config/nodes/solaris-10-sparc.yaml
@@ -1,0 +1,6 @@
+---
+HOSTS:
+  solaris-10-sparc:
+    roles:
+      - agent
+    platform: solaris-10-sparc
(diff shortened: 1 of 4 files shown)


johnduarte:
Solaris yaml files updated per comments from @joshcooper and @MikaelSmith

joshcooper:
@MikaelSmith @johnduarte I was wondering about the yaml triple dash, and why it worked without it when testing solaris 11. From http://yaml4r.sourceforge.net/doc/page/basic_types_in_yaml.htm

> The triple dash '---' in the above outputs is the YAML separator. Documents which contain a single collection (such as the Collections examples on the last page) don't require a separator.

So it works without triple dash, because beaker loads the YAML as a single Hash. That said, it's probably best to be explicit and in the dash now, so :+1:

PR #298, opened 2015-09-07
(HI-466) Added memcache caching method for Hiera backends

brutus333:
I found very useful to have a memcache cache (similar with Filecache class) for custom hiera backends. It seems that using memcache decreases the catalog compilation time three times in our case.
Since this is a useful feature across backends, I think that it should be put in the hiera library. That's why I'm proposing this PR.

On the other hand, the dependency on ruby Dalli class should be evaluated - no official RPM exists for RH and CentOS, only for Fedora and Ubuntu.

Several usage information:

The following code should be specified in the custom backend:

in initialize method:

@cache = cache || Memcache.new(@servers)

(of course, the @servers array should be taken from the configuration file during initialization)

in lookup method:

@cache.fetch(key)
{ search_method(key) }

where key is the lookup key (or something obtained from lookup key) and search_method is the expensive code called for finding the information pertaining to lookup key.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ group :development, :test do
   gem 'mocha', "~> 0.10.5", :require => false
   gem 'json', "~> 1.7", :require => false, :platforms => :ruby
   gem "yarjuf", "~> 1.0"
+  gem 'dalli', "~> 2.7.4"
 end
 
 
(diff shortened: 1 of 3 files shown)


hlindberg:
closed because ticket was closed with "Won't Do". See suggestions on ticket HI 466.

PR #299, opened 2015-09-08
Update hiera to rspec 3.

kylog:


joshcooper:
@kylog do you want to target stable so that future stable->master merges are clean?

kylog:
Good call. #306.

PR #304, opened 2015-09-10
(maint) Add ip and vmhostname keys to SPARC nodes

whopper:
[skip-ci]


whopper:
Merge after unfreeze.

PR #306, opened 2015-09-16
(maint) Update hiera to rspec3

kylog:


spec/unit/config_spec.rb
```diff
@@ -12,10 +12,6 @@ class Hiera
         }
       end
 
-      it "should treat string sources as a filename" do
-        expect { Config.load("/nonexisting") }.to raise_error
-      end
-
```
joshcooper: looks like this is tested in https://github.com/kylog/hiera/blob/maint/stable/rspec3/spec/unit/config_spec.rb#L19, so makes sense to delete this one.

PR #307, opened 2015-09-17
(maint) Improve error message in read_file for filecache

haus:
Previously if a type error occurred during the file_read, the error
message would always say that the data retrieved is a string, and not
the expected type. This is because data comes from File.read, which
returns a string. Having the class of the data from the @cache object
allows the error message to be more helpful and return the class of the
parsed data instead of just 'String'.


branan:
Since @haus poked me... AFAICT this is sane, but would be nice to have feedback from language before I hit the button. @HAIL9000?

HAIL9000:
thumbs-up-computer-kid-gif_348542

branan:
Just waiting for travis

PR #314, opened 2015-10-01
(HI-471) Allow backends to perform qualified key lookup

mattaezell:
Previously, Hiera performed qualified key lookup by asking the backend for the
top-level key and descending into the data.  As an optimization, provide the
fully qualified key to the backend so that it can minimize processing required
to return a large data structure when only a small portion of it is used.  This
can speed up lookups for backends where processing data is expensive (like
eyaml).

### DIFF
--- a/lib/hiera/backend.rb
+++ b/lib/hiera/backend.rb
@@ -263,8 +263,14 @@ class Hiera
             backend = (@backends[backend] ||= find_backend(backend_constant))
             found_in_backend = false
             new_answer = catch(:no_such_key) do
-              value = backend.lookup(segments[0], scope, order_override, resolution_type, context)
-              value = qualified_lookup(subsegments, value) unless subsegments.nil?
+              if subsegments.nil? 
+                value = backend.lookup(key, scope, order_override, resolution_type, context)
+              elsif backend.respond_to?(:lookup_with_segments)
+                value = backend.lookup_with_segments(segments, scope, order_override, resolution_type, context)
+              else
+                value = backend.lookup(segments[0], scope, order_override, resolution_type, context)
+                value = qualified_lookup(subsegments, value) unless subsegments.nil?
+              end
               found_in_backend = true
               value
             end


hlindberg:
@thallgren can you take a look at this? Something we should also consider for the data provider API for puppet 4.3.0.

thallgren:
We discussed this earlier and concluded that dotted names was part of the Hiera specific interpolated expression evaluation and not something we'd like to expose in the overall data provider API where we use `::` to represent qualified names.

Perhaps we need an API where it's up to the data provider to parse the key into segments?

mattaezell:
@thallgren recommended that we define a `qualified_lookup` instance method to `Backend`.  Since `Backend` is a module, its instance methods aren't available to the backend classes unless they `include Backend` (note: I'm not that familiar with Ruby, so if there's a better way to do this please let me know).  Backends don't `include Backend` today, so it would be a breaking change.

Instead, just check if the backend `respond_to?(:qualified_lookup)` otherwise call `lookup`

hlindberg:
actually the term "qualified" is problematic here since that is the term used for qualified names like a::b::c in puppet and with that in mind, every lookup is a lookup of a (potentially) fully qualified name/key. The dot notation following the lookup is actually a projection of the value. I can imagine a method `supports_projection?` that if implemented is called, and if if returning true, a lookup must process a projection expression (segments) before returning. If not supporting this, it is the responsibility of the caller (i.e. "hiera") to perform the projection instead of the backend.

thallgren:
I think it's confusing to talk about 'qualified names' versus 'projection'. It's a name that is split into segments in order to qualify some data in a collection regardless of if that collection is a set of modules, a set of keys in a module, or a set of keys in a hash. How the data is constituted is IMO an implementation detail that should be hidden by the data provider.

Why not instead say that we allow both notations (although not mixed) and forget about if it means qualified or projected? Or alternatively say that we allow the dot notation in Hiera but consider that legacy and deprecated in favor of the '::' notation.

Or can you think of any situation where the 'projection' versus 'qualification' would be semantically different in a meaningful way? I can see them giving different behavior when it comes to merging strategies but enforcing those differences instead of unifying the behavior would introduce complexities that I find it hard to believe that anyone would want to deal with.

hlindberg:
There is quite a big difference between `a::b::c` and `a::b.c`. The name space separator :: is simply different from the '.' (member/part) operator - they can not be interchanged. And yes this is semantically different, `a::b::c` would be the value for the parameter `c` for the class `a::b`, whereas `a::b.c` would be the `c` key in the hash bound to the parameter `b` for class `a`.

```
class a ($b, $c) { }
class a::b($c) { }
```

Whatever we end up calling the notion on "drilling down into the returned value" for a key, it is an operation that not all hiera backends will support. For the data provider API we can simply build it in.

What I don't want is to confuse the Puppet notion of QualifiedName (which does not include internal navigation in attributes), with the concept in lookup that provides this. I do not care much what we call it as long as it is no confusing. (I think 'projection' is a good term for this as it produces a projection of the key's value instead of the entire value).

thallgren:
Your example shows that we must have different separators since we don't want `a($b)` to find the hash containing parameter values for `a::b`. But do we need different terminology?

I still think of this as a qualification, not a projection, in that the intent is to navigate further into some detail of some contextual data. This PR is about having the backend itself do that (and possibly our data providers too) which invalidates the notion of "drilling down into the returned value". It's no longer a "value". It has become a naming context where values can be found.

How about doing what I suggested in my initial comment on this PR (and some other well-known implementations do) and let the backend worry about both parsing and drilling? And let the terminology stay the same.

hlindberg:
Sure, I can see supporting `lookup('a::b.c')` (which does not work now IIRC) as opposed to the equivalent `lookup('a::b')['c']`. For data providers we can simply specify the API to support this (never mind what we call the different parts of the lookup 'expression'. I do agree that the more that is specified, the more it is "qualified", I just think it is somewhat confusing. How about calling it a "lookup expression", which may consist of a "simple key (a qualified name)", or "a a key and a navigation expression into the value bound to that key".

mattaezell:
I just updated the PR - I updated all references to "qualified keys" to be "subkeys".  Backends can now define a `supports_subkeys?` method.  If that returns true, the key will be passed to the backend unchanged.  Otherwise, previous behavior is maintained (hiera handles the subkey lookup).

Is this a viable compromise?

thallgren:
I think this commit was a bit too intrusive in that it changes the name of the method 'qualified_lookup' which in turn changes API (it's not a private method and for all we know, custom backends may call back into this method).

The proposed implementation also queries if there's support for subkeys even when no subkey is present and if a subkey is present, it requires the backend to parse the key into segments again.

mattaezell:
I made all the changes because of the ambiguity that @hlindberg pointed out around the word qualified.  I'm happy to revert them as long as everyone agrees that's the best course forward.

We also have the problem that without changing the API to provide pre-parsed segments, the backends are always going to have to re-parse the key into segments.  Originally I tried stuffing that into `context`, but that is problematic for recursive interpolation situations.  And due to this, the `lookup` method has no knowledge of whether or not Hiera determined that there were subkeys or not.  I thought about saving some state when `supports_subkeys` is called, but there's no good way to reset that state when non-subkey requests are issued.  So a backend that supports subkeys pretty much always has to check and re-parse.

That being said, if it's worthwhile for a backed to support subkey lookups, it's probably relatively trivial to re-parse the subkey compared to looking up and processing extraneous data.

thallgren:
Not sure I understand the reasoning behind why the split must be repeated. Why not do something like this?

    if subsegments.nil?
      value = backend.lookup(key, scope, order_override, resolution_type, context)
    elsif backend.respond_to(:lookup_with_segments)
      value = backend.lookup_with_segments(segments, scope, order_override, resolution_type,context)
    else
      value = backend.lookup(segments[0], scope, order_override, resolution_type, context)
      value = qualified_lookup(subsegments, value)
    end

mattaezell:
I'm fine with that approach.  PR updated.  Thanks.

PR #315, opened 2015-10-08
(HI-464)  Add beaker node defs for Win10(x86,x64)

johnduarte:
[skip-ci]

### DIFF
--- a/acceptance/config/nodes/win10-rubyx64.yaml
+++ b/acceptance/config/nodes/win10-rubyx64.yaml
@@ -1,0 +1,14 @@
+---
+HOSTS:
+  agent-10-x86_64-rubyx64:
+    roles:
+    - agent
+    platform: windows-10-64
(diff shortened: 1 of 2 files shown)


whopper:
@johnduarte we think this needs a rebase to make the travis commit test pass.

johnduarte:
thanks @whopper . rebased and waiting for another travis run.

whopper:
Hm, that didn't work. So Travis is mad because of those duplicate merges in master/stable of the ubuntu node def removal with the `(PUP-...)` tag. Any PR against stable is going to fail on that check now. We're going to need to do something about that to get it out of the git history :( /cc @kylog 

(We'll have the same problem for facter)

whopper:
Note: needs a merge up to master when merged.

kylog:
Merged up to master too.

PR #318, opened 2015-10-21
(HI-475)Use syck YAML parser for Ruby 1.9.3 and 2.0.x

mattymo:
Default YAML parser treats unquoted strings starting
with 0 sometimes as a numeric value.

Example input:
password: 0Xmy_secret_pass
mac: 08:00:27:32:56:08

1.93 parser output:
$ hiera password
0Xmy_secret_pass
$ hiera mac
6226750568

2.0.0 parser output:
$ hiera password
12380245
$ hiera mac
1960347

Output is as expected with 'syck' yamler in 1.9.3 and 2.0.0:
$ hiera password
0Xmy_secret_pass
$ hiera mac
08:00:27:32:56:08

Ruby versions below 1.9.3 do not contain 'syck' yamler and 2.1 is not affected.


MikaelSmith:
Thanks for filing a ticket and working on this issue!

First, please follow the [contributor guidelines](https://github.com/puppetlabs/hiera/blob/master/CONTRIBUTING.md) for commit messages. We'd also prefer not to have the merge commit, so you should rebase your branch to clean up the commits.

Second, this PR really needs some spec tests with it. If you look at the spec run in Jenkins for Ruby 2.0.0, you'll see warnings like `syck has been removed, psych is used instead`, pointing out that [Syck was removed in Ruby 2.0.0](http://ruby-doc.org/stdlib-1.9.3/libdoc/yaml/rdoc/YAML.html#module-YAML-label-Syck). The failure with Ruby 2.1.6 is showing that the commit messages need updating. The error in Ruby 1.9.3 (where Syck is actually used) looks like Syck is messing with rspec's ability to mock YAML calls.

Overall this is a problematic PR, as it will change the behavior of YAML for the Ruby interpreter, impacting Puppet and Facter's YAML engine as well. Seeing as Syck was considered unmaintained in 1.9.3, this needs significant testing. Based on @haus's feedback, using Syck everywhere in Ruby 1.9.3 will probably break things.

Our recommendation is either quote the strings, find a way to patch Ruby to fix Psych, or use a newer version of Ruby.

PR #321, opened 2015-11-02
(HI-431) Update pre-suite for none hypervisor

joshcooper:
This commit removes conditional logic for solaris when the
hypervisor is none. The details of expunging puppet from the
host is delegated to the beaker `remove_puppet_on` method.

This commit was originally made to master, backporting to stable
for HI 3.0.5 release.


johnduarte:
:+1:  this will do it. Thanks @joshcooper

PR #322, opened 2015-11-04
(HI-443) Update platform field to use version numbers

joshcooper:
Previously, beaker tried to install puppet-agent on OSX 10.11 twice
because the 'el' regex matched the elcapitan platform. When the 'el'
regex was made to be more specific, beaker added PermitUserEnvironment
to the pre 10.11 location, because beaker expects platform to contain
version numbers, not codenames.

This commit changes all OSX host configs to version numbers as
codenames, and makes the 'el' regex more specific, as 'el' could occur
in a future OS name.

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -7,7 +7,7 @@ module Puppet
   module Acceptance
     module InstallUtils
       PLATFORM_PATTERNS = {
-        :redhat        => /fedora|el|centos/,
+        :redhat        => /fedora|el-|centos/,
         :debian        => /debian|ubuntu/,
         :debian_ruby18 => /debian|ubuntu-lucid|ubuntu-precise/,
         :solaris       => /solaris/,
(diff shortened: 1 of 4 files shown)


joshcooper:
Passed adhoc pipeline on 10.9 - 10.11

https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Manual/view/Puppet%20Agent%20Ad-Hoc/job/platform_aio-suite_intn-sys_adhoc-hiera/65/

acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -7,7 +7,7 @@ module Puppet
   module Acceptance
     module InstallUtils
       PLATFORM_PATTERNS = {
-        :redhat        => /fedora|el|centos/,
+        :redhat        => /fedora|el-|centos/,
```
MikaelSmith: Those this looks like it's no longer a problem.

joshcooper: yep, but I don't want it to be a problem in the future, if we happen to add another OS whose name (but not version) contains `el`

PR #324, opened 2015-11-24
Added compatibility mode option for deep and deeper merge behavior

Mike-Petersen:
We just created a project that uses hiera and active record. Active support is included automatically and also provides deep_merge. This deep merge does not allow additional options and is in that way not compatible. The deep_merge gem has a compatiblity mode for this, see https://github.com/danielsdeleo/deep_merge#using-deep_merge-in-rails

This commit adds the possibility to use the compatibility mode.

### DIFF
--- a/lib/hiera/backend.rb
+++ b/lib/hiera/backend.rb
@@ -183,13 +183,16 @@ class Hiera
       # Merges two hashes answers with the given or configured merge behavior. Behavior can be given
       # by passing _resolution_type_ as a Hash
       #
-      #  :merge_behavior: {:native|:deep|:deeper}
+      #  :merge_behavior: {:native|:deep|:deeper|:deep_compat|:deeper_compat}
       #
       # Deep merge options use the Hash utility function provided by [deep_merge](https://github.com/danielsdeleo/deep_merge)
+      # If you also include active support you need a compatibility mode from [deep_merge](https://github.com/danielsdeleo/deep_merge#using-deep_merge-in-rails)
       #
-      #  :native => Native Hash.merge
-      #  :deep   => Use Hash.deep_merge
-      #  :deeper => Use Hash.deep_merge!
+      #  :native        => Native Hash.merge
+      #  :deep          => Use Hash.deep_merge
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This looks great, thanks for the contribution.

hlindberg:
I do not like this at all since it exposes the alternative methods to the user. This is not something we want to maintain. I am therefore reverting this PR.

I can accept that the implementation adapts to what is actually installed. Also note that if rails is used with Puppet the limits it imposes will also affect "data in modules" and lookup.

hlindberg:
Also note that commits should be written in imperative form; "Add" instead of "Added" in this case, and that the PR should be marked with the ticket number.

MikaelSmith:
So the suggestion would be to instead overload deep_merge? I'm not clear what you mean by "I can accept that the implementation adapts to what is actually installed."

hlindberg:
I mean: either always use the `compat` versions if they produce exactly the same result, or only use them if they are available (if there is something special that is needed to make them available). There should be no need for the user to know and set those options. It is always just `deep` or `deeper` and the options (in our API) should always be available.

If we do not do it that way - the eco-system will break down because options that users have crafted using the new lookup_options would otherwise in some configurations not work.

(We need to make the same changes in Puppet as well if Puppet is ever combined with Rails).

Mike-Petersen:
I changed the implementation to use the compat mode as default. But i do not know if i can add this commit to this pull request. Can you give me a hint?

hlindberg:
depends on what you did; if you made additional changes on the same branch, or created a new branch. If you continued on the same branch, then rebase it interactively and remove the two prior commits, then do a forced push to github. The PR will then automatically be updated.

If you created a new branch. Delete the old locally. Rename the new to the name of the deleted. Force push that branch to github and the PR updates automatically.

Or - just close this PR and make a new for your new branch.

Does that help?

(And thanks for making those changes)

Mike-Petersen:
I rebased it interactively and force pushed it to github. But it seems to not update the merge request. I have created a new pull request here: https://github.com/puppetlabs/hiera/pull/326

PR #326, opened 2015-11-27
(HI-483) Use compaitibility mode from deep_merge as default

Mike-Petersen:
With this commit you can use hiera and active support in the same project.

See discussion in previous pull request https://github.com/puppetlabs/hiera/pull/324

### DIFF
--- a/lib/hiera/backend.rb
+++ b/lib/hiera/backend.rb
@@ -2,7 +2,7 @@ require 'hiera/util'
 require 'hiera/interpolate'
 
 begin
-  require 'deep_merge'
+  require 'deep_merge/rails_compat'
 rescue LoadError
 end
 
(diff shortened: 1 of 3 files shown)


MikaelSmith:
`rails_compat` seems to be part of deep_merge. That's pretty terrible.

@hlindberg does this seem reasonable, or did you have another approach in mind?

hlindberg:
I am fine with this approach as long as it is 100% the same as doing the non compat, and that it does not clash with puppet and hiera being used together. I hope that reference to "rails" in the require does not imply in any way that anything "rails" related must be present.

MikaelSmith:
It requires https://github.com/danielsdeleo/deep_merge/blob/master/lib/deep_merge/rails_compat.rb, which doesn't require anything except `deep_merge/core`.

hlindberg:
Thanks @MikaelSmith that looks fine - poor choice on  naming the feature something with "rails" although I understand why.

PR #328, opened 2015-12-14
(HI-220) Deprecate use of interpolation methods in Hiera config

thallgren:
Using interpolation methods in the hiera configuration configuration
file should be avoided. The 'hiera' method in particular will lead
to problems since it's very likely that it will cause an endless
recursion when a new attempt is made to read the configuration file.

This commit ensures that a deprecation warning is issued when the
interpolation method syntax is used in a hiera configuration file. An
error will be raised in case the 'hiera' interpolation method is used.

### DIFF
--- a/lib/hiera/interpolate.rb
+++ b/lib/hiera/interpolate.rb
@@ -9,7 +9,12 @@ class Hiera::Interpolate
     INTERPOLATION = /%\{([^\}]*)\}/
     METHOD_INTERPOLATION = /%\{(scope|hiera|literal|alias)\(['"]([^"']*)["']\)\}/
 
-    def interpolate(data, scope, extra_data, context)
+
+    def interpolate_config(data, scope, context)
+      interpolate(data, scope, {}, context, true)
+    end
+
+    def interpolate(data, scope, extra_data, context, is_config = false)
       if data.is_a?(String)
         # Wrapping do_interpolation in a gsub block ensures we process
         # each interpolation site in isolation using separate recursion guards.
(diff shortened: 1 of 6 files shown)


lib/hiera/interpolate.rb
```diff
@@ -41,17 +46,17 @@ def interpolate(data, scope, extra_data, context)
       end
     end
 
-    def do_interpolation(data, scope, extra_data, context)
+    def do_interpolation(data, scope, extra_data, context, is_config)
```
hlindberg: is_config needs to have default value or it breaks API

lib/hiera/interpolate.rb
```diff
@@ -59,9 +64,19 @@ def do_interpolation(data, scope, extra_data, context)
     end
     private :do_interpolation
 
-    def get_interpolation_method_and_key(data)
+    def get_interpolation_method_and_key(data, is_config)
```
hlindberg: same here - default value needed to be API compliant

lib/hiera/interpolate.rb
```diff
@@ -59,9 +64,19 @@ def do_interpolation(data, scope, extra_data, context)
...
       if (match = data.match(METHOD_INTERPOLATION))
-        case match[1]
+        key = match[1]
+        if(is_config)
+          if key == 'hiera'
+            # Letting this pass while parsing the config is very likely to result in another attempt to parse
+            # the config and hence, endless recursion.
+            raise Hiera::InterpolationInvalidValue, "Cannot use interpolation method 'hiera' in hiera configuration file"
```
hlindberg: This is a breaking change for those than managed to configure their hiera carefully to make it work. This will require them to make changes. To make it an error we need a major HI release, which we do not want. Suggest just making it be a deprecation warning.

lib/hiera/interpolate.rb
thallgren: Please note that both `do_interpolation` and `get_interpolation_method_and_key` are private methods.

lib/hiera/interpolate.rb
thallgren: I've tried but not succeeded in creating such a configuration (tested among other things the example in the JIRA). I don't see how it can succeed given that any hiera lookup will call datasources which in turn will call back to interpolate.

hlindberg: IIRC someone posted an example they claimed worked. It is still wonky. I am inclined to just error out though.

hlindberg:
ok, questions answered. And the use of recursive configuration is so strange that it should have been an error from the start. Merging.

PR #329, opened 2015-12-14
Revert "(HI-220) Deprecate use of interpolation methods in Hiera config"

thallgren:
Reverts puppetlabs/hiera#328


thallgren:
I have a better solution that permits 'hiera' and detects lookup recursion.

PR #335, opened 2015-12-30
(HI-486) Bump to beaker-hostgenerator 0.3.1

kylog:
The puppet-agent components need the osinfo version 1 of b-hg,
so that centos4 is correctly specified as a platform name, so
that puppet-agent can be installed on centos4.

This commit moves to b-hg 0.3.1, the first version implementing
that, and specifies the osinfo version when invoking b-hg.


waynr:
:+1:

PR #336, opened 2016-01-06
(HI-488) Add Arista Support

LuvCurves:


joshcooper:
Can you rebase (it looks like there are conflicts)? Also it looks like the setup step is missing from the commit.

PR #338, opened 2016-01-11
(maint) Acceptance Rake - give CONFIG priority

johnduarte:
This commit changes the logic in the beaker_test rake task for
acceptance giving BEAKER_HOSTCONFIG priority over TEST_TARGET.
It also changes the name of the ENV variable from CONFIG to
BEAKER_HOSTCONFIG

In the case that a static beaker host config file is needed for
running acceptance testing, the file path to this config should
be assigned to the ENV variable BEAKER_HOSTCONFIG. This value
will take precedence over any value set for TEST_TARGET.

TEST_TARGET is still used to supply a string for dynamically
generating beaker host configs.

skip:ci

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -95,15 +95,15 @@ EOS
   tests_opt = "--tests=#{tests}" if tests
 
   target = ENV['TEST_TARGET']
-  if target
+  if config and File.exists?(config)
+    config_opt = "--hosts=#{config}"
+  elsif target
     cli = BeakerHostGenerator::CLI.new([target, '--disable-default-role', '--osinfo-version', '1'])
-    ENV['CONFIG'] = "tmp/#{target}-#{SecureRandom.uuid}.yaml"
+    ENV['BEAKER_HOSTS'] = "tmp/#{target}-#{SecureRandom.uuid}.yaml"
     FileUtils.mkdir_p('tmp')
     File.open(config, 'w') do |fh|
       fh.print(cli.execute)
     end
-    config_opt = "--hosts=#{config}"
-  elsif config
     config_opt = "--hosts=#{config}"
   end
 
@@ -244,7 +244,7 @@ def sha
 end
 
 def config
-  ENV['CONFIG']
+  ENV['BEAKER_HOSTS']
 end
 
 namespace :ci do
@@ -254,7 +254,7 @@ namespace :ci do
       sh("beaker",
           "--type", "git",
           "--load-path", "lib",
-          "--hosts", ENV['CONFIG'],
+          "--hosts", ENV['BEAKER_HOSTS'],
           "--pre-suite", "setup/git",
           "--install", "PUPPET/#{ENV['PUPPET'] || 'master'},FACTER/#{ENV['FACTER'] || 'master'},HIERA/#{ENV['HIERA'] || 'master'}",
           "--tests", "tests",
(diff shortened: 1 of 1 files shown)


joshcooper:
`BEAKER_HOSTCONFIG` almost sounds like it should be a beaker-hostgenerator string? we could do `BEAKER_HOSTCONFIG_FILE`, but I'm also fine as is too.

nit, travis failed because of QENG in the summary. Can you change to (maint)?

johnduarte:
@joshcooper 
Having chatted with @er0ck and @waynr , future intent would be to have `BEAKER_HOSTCONFG` become an ENV var that `beaker` would consume that would take precedent over the `--config` flag. In that context, I think it make sense without the `_FILE` suffix.

justinstoller:
I don't like the idea of having an environment variable (something that can be set per session) override a command line flag (something that's set per invocation). It seems like having a less specific input override a more specific input.

er0ck:
i hate to keep this conversation going longer, buuut:
i believe --config is deprecated in favor of --hosts.
in that context, should we use BEAKER_HOSTS?
in almost all cases beaker's environment variables map directly to the options' spelling with a pre-pended BEAKER_

johnduarte:
@justinstoller your order of precedence is probably correct. I am sure I just misspoke with regard to the intent of the override.

@er0ck you are correct about the deprecation of `--config`. In that context, I agree that `BEAKER_HOSTS` makes more sense.

er0ck:
@justinstoller IMO, you are correct when talking about a single application and not an abstraction.  in this case i believe we're talking about using ENV vars as args to rake which is abstracting a small portion of how to run <testharness>.
args to rake tasks are hard to handle, in some cases, and are only useful if defined correctly by a rake task *author*.
so i think wayne and others mean that a rake file *user* should be able to modify important portions of a raketask using environment variables to modify how a task runs beaker.  i'd also be okay with arguments in some cases.  in this case it makes sense to use a beaker ENV and just forward it on for the most transparency, discoverability.
we also have precedence with CONFIG and SHA.

i hope i didn't misinterpret your concern...

er0ck:
oh.  i totally misinterpreted your concern @justinstoller.  yes, i think rake would  just forward on any BEAKER_* ENV and they'd be combined with that task's (or child tasks?) beaker cli flags and do_the_right_thing :tm:  (beaker flags take precedence over env)

justinstoller:
:+1:

johnduarte:
@joshcooper  or @kylog if this looks  alright, I can reproduce for the other agent components.

joshcooper:
:+1:

johnduarte:
I forgot to verify that the `BEAKER_HOSTS` is a file. Let me patch it one more time.

joshcooper:
Tested both:

```
$ bundle exec rake ci:test:aio BEAKER_HOSTS=ubuntu.yaml SHA=5e27c223c410a4d41de2716315e8682156278a80 SUITE_VERSION=1.3.2.203.g5e27c22 TESTS=tests/yaml_backend/00-setup.rb
```

```
bundle exec rake ci:test:aio TEST_TARGET=ubuntu1510-64a SHA=5e27c223c410a4d41de2716315e8682156278a80 SUITE_VERSION=1.3.2.203.g5e27c22 TESTS=tests/yaml_backend/00-setup.rb
```

joshcooper:
I merged this up to master too

PR #340, opened 2016-01-25
(HI-943) Update hiera.bat to current standard

melissa:
With the switch over to building puppet-agent for windows with Vanagon,
we need to effectively install hiera with install.rb. This script
installs hiera.bat, which is a wrapper that lets windows handle ruby
scripts. This commit updates hiera.bat to use our environment.bat file if
we install it, and to be more in line with the hiera.bat file installed
with the puppet-agent package as built by puppet_for_the_win

This takes advantage of `environment.bat` in puppet-agent installed with https://github.com/puppetlabs/puppet-agent/pull/512

### DIFF
--- a/install.rb
+++ b/install.rb
@@ -255,10 +255,13 @@ def install_binfile(from, op_file, target)
       tmp_file2 = Tempfile.new('hiera-wrapper')
       cwv = <<-EOS
 @echo off
-setlocal
-set RUBY_BIN=%~dp0
-set RUBY_BIN=%RUBY_BIN:\\=/%
-"%RUBY_BIN%ruby.exe" -x "%RUBY_BIN%hiera" %*
+SETLOCAL
+if exist "%~dp0environment.bat" (
+  call "%~dp0environment.bat" %0 %*
+) else (
+  SET "PATH=%~dp0;%PATH%"
+)
+ruby.exe -S -- hiera %*
 EOS
       File.open(tmp_file2.path, "w") { |cw| cw.puts cwv }
       install(tmp_file2.path, File.join(target, "#{op_file}.bat"), :mode => 0755, :preserve => true, :verbose => true)


melissa:
Still testing

install.rb
```diff
@@ -255,10 +255,13 @@ def install_binfile(from, op_file, target)
...
-set RUBY_BIN=%~dp0
-set RUBY_BIN=%RUBY_BIN:\\=/%
-"%RUBY_BIN%ruby.exe" -x "%RUBY_BIN%hiera" %*
+SETLOCAL
+if exist "%~dp0environment.bat" (
+  call "%~dp0environment.bat" %0 %*
+) else (
+  SET PATH=%~dp0;%PATH%
+)
+ruby.exe -S -- hiera %*
```
shrug: Which ruby gets called here? Does it set the path to the bundled interpreter?

melissa: It should be the ruby that's on the path. Which, if this is called as a part of the puppet-agent package, it should be finding our ruby first, since we set our bindir to be the first thing on the path https://github.com/puppetlabs/puppet-agent/pull/512/files#diff-f19be7ccad3d10336f31489de24d09ddR20

shrug:
:+1: , short of explicitly using the fully qualified path to the binaries, which may be unecessary

melissa:
cc/ @MikaelSmith @kylog @joshcooper could one of you review this?

joshcooper:
I'm assuming we know `hiera.bat` is installed into `<INSTALLDIR>\puppet\bin\hiera.bat` and we know the path to our ruby is `<INSTALLDIR>\bin\ruby.exe`, so we know the relative path to ruby from `hiera.bat` is something like `%~dp0\..\..\bin\ruby.exe`? Can we always use that instead of relying on `PATH`?

Is there a case where `environment.bat` won't exist?

melissa:
@joshcooper the problem is we don't always know where ruby will be. Since this is the code that gets installed when someone runs `install.rb` on their system, we have no idea what ruby they might be using. If it's in puppet-agent, then we know ruby.exe will be in the same directory as hiera.bat (assuming that layout change gets the okay). Again, in the puppet-agent case, we always know where ruby is and where environment.bat is. If someone is using install.rb out in the wild, there are no guarentees of those files.

For anyone interested, the `hiera.bat` installed from puppet_for_the_win is
```

@echo off
SETLOCAL

call "%~dp0environment.bat" %0 %*

ruby -S -- "%SCRIPT_NAME%" %*
```

install.rb
glennsarti: @melissa Does it matter that the default search order for executables is to look in the current working directory and THEN the path?  Conceivably it could be possible to have a different ruby.exe in the local dir but I think this might be too unlikely to happen.

What do you think?

melissa: I'm not sure. I don't think so. Originally, this batch file assumes ruby is in the current directory. It's not obvious, but environment.bat installed in puppet-agent just sets path to include the current directory where ruby is. It's mainly set up like this to preserve functionality....

MikaelSmith:
Looks good in stand-alone install and a puppet-agent build. :+1:

MikaelSmith:
Oops, that had the wrong ticket number.

melissa:
D'oh!! Sorry about that :(

PR #342, opened 2016-02-03
(HI-494) Bring back support for empty interpolation

hlindberg:
Before this, we started doing something wrong in the hiera 3.x series as
interpolation of empty strings 

See commit for details.


hlindberg:
I changed the gsub to a strip.  A strip! cannot be used as that would alter the value, if it is not empty after space removal, it is used in original form. It is (i.e. I am) not 100% certain that whitespaces are not significant elsewhere.

PR #343, opened 2016-02-23
(maint) Fix erroneous use of .value in README

pizzaops:
Calling `.values` on the loaded facts scope breaks the example. Not sure
why that's there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ require 'hiera'
 require 'puppet'
 
 # load the facts for example.com
-scope = YAML.load_file("/opt/puppetlabs/puppet/cache/yaml/facts/example.com.yaml").values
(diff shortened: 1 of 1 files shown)


hlindberg:
Thank you for the contribution.

hlindberg:
As you can see, the tests failed because the commit is not following the standard form. You do not have to open a ticket for this minor problem, Could you update the commit to say "(maint) Fix erroneous use of .value in README", and I will get this merged. If this is too much of a hassle for you, let me know and will make the change for you.

pizzaops:
@hlindberg fixed the commit title and the pr title (I think?)

hlindberg:
Thanks @pizzaops

PR #344, opened 2016-02-25
(HI-499) Fail if backend cannot be loaded

hlindberg:
Before this, if a backend could not be loaded, this was only logged as a
warning and hiera merrily continued by serving data from remaining (if
any) backends.

This can lead to catastrophic results as lookups may be resolved with
large portions of the data missing. This in turn can cause catalogs to
contain a radically different configuration causing a wide range of
unpleasant surprises.

### DIFF
--- a/lib/hiera/config.rb
+++ b/lib/hiera/config.rb
@@ -74,7 +74,7 @@ class Hiera::Config
         begin
           require "hiera/backend/#{backend.downcase}_backend"
         rescue LoadError => e
-          Hiera.warn "Cannot load backend #{backend}: #{e}"
+          fail "Cannot load backend #{backend}: #{e}"
         end
       end
     end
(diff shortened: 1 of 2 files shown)


lib/hiera/config.rb
```diff
@@ -70,11 +70,12 @@ def yaml_load_file(source)
     private :yaml_load_file
 
     def load_backends
+      puts "LOADING BACKENDS"
```
kylog: Remove the `puts`?

hlindberg: doh

lib/hiera/config.rb
```diff
@@ -74,7 +74,7 @@ def load_backends
         begin
           require "hiera/backend/#{backend.downcase}_backend"
         rescue LoadError => e
-          Hiera.warn "Cannot load backend #{backend}: #{e}"
+          fail "Cannot load backend #{backend}: #{e}"
```
kylog: Was using `fail` vs `raise` here a conscious decision? Fwiw I don't see any other uses of `fail` in hiera.

PR #345, opened 2016-03-17
(HI-504) Add ability to use quoted keys

thallgren:
This commit adds the ability to put quotes around keys. A key can be
quoted with either single or double quotes. A quoted key will be used
verbatim instead of splitting it on dots. In other words, quoting a key
will not use dots to navigate into hashes and arrays. Instead it will
match a dotted key at top level.

Interpolation expressions that use interpolation methods must have the
key quoted twice to get the effect. The reason for this is that the
argument to an interpolation method already had to be quoted. In order
to not introduce incompatibilities, a quoted key in this context must
therefore be quoted twice. Example:

```
---
a.quoted.key: 'Value for a quoted key'
ref.to.key: 'Put the value here: %{hiera("''a.quoted.key''")}'
```


hlindberg:
The most common thing is to quote a portion of the key: i.e
```
a.'verbatim.key'
```
Using the same rules as in Facter

hlindberg:
Is the double single quotes needed because of the Yaml escape rules? Would this be the same:
```
ref.to.key: 'Put the value here: %{hiera("\'a.quoted.key\'")}'
```
As per this spec: http://yaml.org/spec/current.html#id2534365

spec/unit/interpolate_spec.rb
```diff
@@ -57,8 +57,90 @@
...
+    it 'should produce an error when different quotes are used on either side in a top-level key' do
+      expect { hiera.lookup("'the.key\"", nil, {}) }.to raise_error(/Unbalanced quotes in key: 'the.key"/)
+    end
+  end
+
+  context 'when doing interpolation with override' do
+    let!(:fixtures) { File.join(HieraSpec::FIXTURE_DIR, 'override') }
 
     it 'should resolve interpolation using the override' do
       expect(hiera.lookup('foo', nil, {}, 'alternate')).to eq('alternate')
```
hlindberg: Test case for partially quoted key required. e.g. that these keys have 4 parts.
```
part1.´verbatim.verbatim´.part3.part4
part1.´verbatim.0´.part3.part4
```

thallgren: Partially quoted keys isn't implemented as it would require a token parser (lexer) to parse instead of using regexps.

hlindberg: yes absolutely, that is the reason this is of value in the first place. Only working on entire key is a small special case.

hlindberg:
needs a rebase, and I would like some input on the questions regarding the double quoting (is it Yaml, and can escapes be used etc.).

thallgren:
The example is YAML. Your example is not valid YAML. You can't escape a single quote with a backslash in a single quoted string. Escaping double quotes within a double string would work though:
```
ref.to.key: "Put the value here: %{hiera('\"a.quoted.key\"')}"
```

README.md
```diff
@@ -101,6 +101,34 @@ $ hiera ssh_users.0
...
+
+<pre>
+a.dotted.key: "data for 'a.dotted.key'"
+other.key: 'here is %{hiera("''a.dotted.key''")}'
+</pre>
+
+Note that two single qoutes are used to escape a single quote inside a single qouted string
+(that's yaml syntax, not Hiera) and that the quoted key must be quoted in turn.
+
+
```
hlindberg: The above needs to be updated showing partial key. That use case is more important and more common than the entire key, so I suggest that is shown first and making "the entire key" a special case.

It would be good to say the examples are Yaml up front. Users reading this will wonder how to do the quoting in Json as well.

thallgren: I think the examples should be YAML only since the idea is to convey the quoting semantics. Users who want to learn YAML/JSON should look at their respective reference (no other examples use JSON).

hlindberg:
After reading the code and updated readme, I can answer my own question about quoting. I left some comments about quoting and the readme in the commits.

hlindberg:
You double quoted example with backslashes is much easier to read and understand. Think it should be part of the readme as a comparisson against the single quoted. (Users, using YAML will not typically know these things.

README.md
```diff
@@ -103,31 +103,45 @@ root
...
+other.key: 'hiera data %{hiera("a.''b.c''.d")}'
 </pre>
 
-Note that two single qoutes are used to escape a single quote inside a single qouted string
-(that's yaml syntax, not Hiera) and that the quoted key must be quoted in turn.
-
+Note that two single quotes are used to escape a single quote inside a single quoted string
+(that's YAML syntax, not Hiera) and that the quoted key must be quoted in turn.
 
 ## Future Enhancements
```
hlindberg: much better! This is readable and understandable

PR #347, opened 2016-03-23
(HI-480) Check for escaped period when determining hash segments

gmarkey:
Likely not the ideal way to solve this problem, but the original code is a bit overzealous with its split.

Performs backward regex lookup to determine if the period has been escaped, then removes escaping.


hlindberg:
HI-504 already has an implementation that allows quoting parts of keys. Dots is not the only problematic character in keys. Will be released in Hiera 3.1.1.

Sorry for not catching your ticket and providing feedback before you went to the trouble of writing this PR.

PR #353, opened 2016-05-03
(HI-510) Look for hiera.yaml in codedir first, then confdir

thallgren:
This commit ensures that when Hiera is initialized, it will first check
if a config file is specified using a CLI option. If it is, then that
config will be used (and an error is raised if it doesn't exist). When
no CLI option is used, Hiera will first look `hiera.yaml` in the
codedir and if no such file exists, it will default to `hiera.yaml`
in confdir.

### DIFF
--- a/bin/hiera
+++ b/bin/hiera
@@ -28,7 +28,7 @@ require 'pp'
 
 options = {
   :default => nil,
-  :config  => File.join(Hiera::Util.config_dir, 'hiera.yaml'),
+  :config  => nil,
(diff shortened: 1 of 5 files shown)


hlindberg:
Looks fine to me. @kylog, can you catch an extra eye on the path changes just to be sure ?

kylog:
Yep, the path changes match up with those in https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md.

PR #357, opened 2016-07-06
(DO NOT MERGE) (HI-341) add native HOCON backend support 

janelu2:
This adds HOCON backend to hiera as well as a HOCON spec test to make sure that it is properly supported.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
 gem "hiera", :path => File.dirname(__FILE__), :require => false
 gem 'deep_merge', :require => false
+gem 'hocon', "~> 1.1.1"
 
 group :development do
   gem 'watchr'
--- a/spec/unit/backend/hocon_backend_spec.rb
+++ b/spec/unit/backend/hocon_backend_spec.rb
@@ -1,0 +1,137 @@
+require 'spec_helper'
+require 'hiera/backend/hocon_backend'
+require 'hocon/config_error'
+
+ConfigWrongTypeError = Hocon::ConfigError::ConfigWrongTypeError
+
+class Hiera
+  module Backend
+    class FakeCache
+      attr_accessor :value
+      def read(path, expected_type, default, &block)
+        read_file(path, expected_type, &block)
+      rescue => e
+        default
+      end
+
+      def read_file(path, expected_type, &block)
+        output = block.call(@value)
+        if !output.is_a? expected_type
+          raise TypeError
+        end
+        output
+      end
+    end
+
+    describe Hocon_backend do
+      before do
+        Config.load({})
+        Hiera.stubs(:debug)
+        Hiera.stubs(:warn)
+        @cache = FakeCache.new
+        @backend = Hocon_backend.new(@cache)
+      end
+
+      describe "#initialize" do
+        it "should announce its creation" do # because other specs checks this
+          Hiera.expects(:debug).with("Hiera HOCON backend starting")
+          Hocon_backend.new
+        end
+      end
+
+      describe "#lookup" do
+        it "should pick data earliest source that has it for priority searches" do
+          Backend.expects(:datasourcefiles).with(:hocon, {}, "hocon", nil).yields(["one", "/nonexisting/one.hocon"])
+          @cache.value = "key: answer"
+
+          expect(@backend.lookup("key", {}, nil, :priority, nil)).to eq("answer")
+        end
+      end
+
+      describe "handling unexpected HOCON values" do
+          before do
+            Backend.expects(:datasourcefiles).with(:hocon, {}, "hocon", nil).yields(["one", "/nonexisting/one.hocon"])
+          end
+
+          it "throws :no_such_key when key is missing in HOCON" do
+            @cache.value = "{}"
+            expect { @backend.lookup("key", {}, nil, :priority, nil) }.to throw_symbol(:no_such_key)
+          end
+
+          it "returns nil when the HOCON value is nil" do
+            @cache.value = "{key: null}"
+            expect(@backend.lookup("key", {}, nil, :priority, nil)).to be_nil
+          end
+
+          it "throws :no_such_key when the HOCON file is false" do
+            @cache.value = ""
+            expect { @backend.lookup("key", {}, nil, :priority, nil) }.to throw_symbol(:no_such_key)
+          end
+
+          it "raises a TypeError when the HOCON value is not a hash" do
+            @cache.value = "[1, 2, 3]"
+            expect { @backend.lookup("key", {}, nil, :priority, nil) }.to raise_error(ConfigWrongTypeError)
+          end
+      end
+
+      it "should build an array of all data sources for array searches" do
+        Backend.expects(:datasourcefiles).with(:hocon, {}, "hocon", nil).multiple_yields(["one", "/nonexisting/one.hocon"], ["two", "/nonexisting/two.hocon"])
+        @cache.expects(:read_file).with("/nonexisting/one.hocon", Hash).returns({"key"=>"answer"})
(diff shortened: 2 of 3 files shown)


janelu2:
@cprice404

hlindberg:
The commits needs to be squashed, and the formalia regarding commits reported as errors by the travis checks should be fixed.

hlindberg:
@kylog ok to add dependency on hocon? (asking becuase of scars from semver and msgpack).

Gemfile
```diff
@@ -2,6 +2,7 @@ source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
 gem "hiera", :path => File.dirname(__FILE__), :require => false
 gem 'deep_merge', :require => false
+gem 'hocon', "~> 1.1.0"
```
cprice404: Needs bump after we do our 1.1.1 release

spec/unit/backend/hocon_backend_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+
+          it "raises a TypeError when the HOCON value is not a hash" do
+            @cache.value = "a: [one, two, three]"
+            expect { @backend.lookup("key", {}, nil, :priority, nil) }.to raise_error(TypeError)
+          end
+      end
+
+      it "should build an array of all data sources for array searches" do
+        Backend.expects(:datasourcefiles).with(:hocon, {}, "hocon", nil).multiple_yields(["one", "/nonexisting/one.hocon"], ["two", "/nonexisting/two.hocon"])
+        @cache.expects(:read_file).with("/nonexisting/one.hocon", Hash).returns({"key"=>"answer"})
```
cprice404: It's possible (likely?) that we'll end up wanting to use the `.conf` file extension for these files, since that's the extension that we use everywhere else that we currently use HOCON.  I added a comment on the Jira ticket to solicit input on that, so let's not change it just yet, but if you wanted to poke around to figure out what it would take, that wouldn't hurt.

spec/unit/backend/hocon_backend_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+        @cache.expects(:read_file).with("/nonexisting/two.hocon", Hash).returns({"key"=>"answer"})
+        expect(@backend.lookup("key", {}, nil, :array, nil)).to eq(["answer", "answer"])
+      end
+
+      it "should ignore empty hash of data sources for hash searches" do
+        Backend.expects(:datasourcefiles).with(:hocon, {}, "hocon", nil).multiple_yields(["one", "/nonexisting/one.hocon"], ["two", "/nonexisting/two.hocon"])
+
+        @cache.expects(:read_file).with("/nonexisting/one.hocon", Hash).returns({})
+        @cache.expects(:read_file).with("/nonexisting/two.hocon", Hash).returns({"key"=>{"a"=>"answer"}})
```
cprice404: These tests are really confusing to me.  It seems like they are mocking `read_file` to make it return a literal Ruby hash.  However, my understanding of the production code is that read_file should return a raw YAML/HOCON string, which the backend should end up parsing via `lookup`.

The YAML tests that you copied this from are doing the same weird mocking here, so this is probably OK/correct, but I'm confused by it and it might be worth digging into just a bit to see if we can better understand what's going on.

spec/unit/backend/hocon_backend_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+      end
+
+      it "should fail when trying to merge an Array" do
+        Backend.expects(:datasourcefiles).with(:hocon, {}, "hocon", nil).multiple_yields(["one", "/nonexisting/one.hocon"], ["two", "/nonexisting/two.hocon"])
+
+        @cache.expects(:read_file).with("/nonexisting/one.hocon", Hash).returns({"key"=>{"a"=>"answer"}})
+        @cache.expects(:read_file).with("/nonexisting/two.hocon", Hash).returns({"key"=>["a", "wrong"]})
+
+        expect { @backend.lookup("key", {}, nil, :hash, nil) }.to raise_error(Exception, "Hiera type mismatch for key 'key': expected Hash and got Array")
+      end
```
cprice404: Since none of the tests above have any HOCON strings in them and don't end up actually parsing any HOCON data, in a way they are effectively duplicating the exact test coverage that is happening in the YAML one.  It might be worth DRY'ing the tests up and introducing some kind of 'common' backend test.  That might not really be worth the ROI, it seems like the JSON specs duplicate some of this too.  Something to think about but probably really necessary for now.

spec/unit/backend/hocon_backend_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+
+      it "should parse the answer for scope variables" do
+        Backend.expects(:datasourcefiles).with(:hocon, {"rspec" => "test"}, "hocon", nil).multiple_yields(["one", "/nonexisting/one.hocon"])
+
+        @cache.expects(:read_file).with("/nonexisting/one.hocon", Hash).returns({"key"=>"test_%{rspec}"})
+
+        expect(@backend.lookup("key", {"rspec" => "test"}, nil, :priority, nil)).to eq("test_test")
+      end
+
+      it "should retain datatypes found in yaml files" do
```
cprice404: s/yaml/hocon

cprice404:
This LGTM, but we have a few things to sort before merging:

1. HOCON release
2. Bump HOCON version here
3. Bump HOCON version in puppet-agent
4. Switch file extension to `.conf` depending on feedback on the jira ticket.
5. Address travis failures and git hygiene issues that @hlindberg pointed out.

Maybe also trying to get a better understanding of how those mocks where `read_file` is returning a ruby literal even work at all :)

Gemfile
kylog: Well, the `~>` should mean it will pick up 1.1.1 (but not 1.2.0), so the bump may not be needed (unless it actually won't work with 1.1.0).

cprice404: Correct, it actually won't work with 1.1.0 because we found a bug.

kylog:
In addition to:
> Bump HOCON version in puppet-agent

We may need to change the puppet-agent component to install to vendor_ruby as discussed in https://groups.google.com/a/puppet.com/forum/?utm_medium=email&utm_source=footer#!msg/language-team/RvEcHDAaIDg/tZV6v_JoAQAJ.

cprice404:
> We may need to change the puppet-agent component to install to vendor_ruby as discussed in https://groups.google.com/a/puppet.com/forum/?utm_medium=email&utm_source=footer#!msg/language-team/RvEcHDAaIDg/tZV6v_JoAQAJ.

Good call - do we have prior art on that, or will we be blazing a trail here?

Gemfile
kylog: Ah cool, got it.

kylog:
>> We may need to change the puppet-agent component to install to vendor_ruby as discussed in https://groups.google.com/a/puppet.com/forum/?utm_medium=email&utm_source=footer#!msg/language-team/RvEcHDAaIDg/tZV6v_JoAQAJ.

> Good call - do we have prior art on that, or will we be blazing a trail here?

You're blazing. I hoped to get to it for semantic_puppet but that intention was overtaken by events.

cprice404:
> You're blazing. I hoped to get to it for semantic_puppet but that intention was overtaken by events.

Dang it.  Events are the worst.

Gemfile
joshcooper: If this is a hard dependency, we'll need to add it to `.gemspec` and `ext/project_data.yaml` as well.

janelu2:
@cprice404 changed all the mock test files to be .hocon as well

joshcooper:
Hiera 5 has builtin support for hocon backends. https://puppet.com/docs/puppet/latest/hiera_config_yaml_5.html#configuring-a-hierarchy-level-built-in-backends. Closing this PR.

PR #358, opened 2016-08-03
(HI-528) Remove json_pure dependency

kylog:
json_pure was added as a dependency to hiera to enable easier development
with ruby 1.8.7 (which doesn't include a json implementation). With hiera
now supporting only ruby 1.9.3+, the dependency is no longer needed.

### DIFF
--- a/.gemspec
+++ b/.gemspec
@@ -29,16 +29,4 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
   s.rubygems_version = "1.8.25"
   s.summary = "Light weight hierarchical data store"
-
-  if s.respond_to? :specification_version then
-    s.specification_version = 3
-
-    if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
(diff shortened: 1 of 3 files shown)


kylog:
Ping @haus whom I asked about this.

Also @cprice404: will this be okay in jruby/puppetserver?

kylog:
Wait, travis 1.9.3 is failing because it's still looking for json_pure. I am confused ....

kylog:
Quoting haus:
looks like jruby brings in the json gem via some process
https://github.com/jruby/jruby/blob/jruby-1_7/lib/pom.rb#L11-L18
so should be fine

haus:
@kylog there's a json entry in the gemfile as well. do you want to do anything with that?

cprice404:
I was going to say what @haus said, I think it should be OK.

kylog:
@haus good catch. I pushed a commit removing that too. Thanks.

haus:
:+1: looks good to me

PR #361, opened 2016-08-26
Allow multiple literals in one line

cwood:
If you try to use literal more than once per line this will fail since it already has been included. I have written something to simple check if it is a literal keyword or not. I am sure this isn't the right way to do this but I want to make you aware of it.

I am using it like the following:

```yaml
hiera::hierarchy:
   - "nodes/%{literal('%')}{::environment}/%{literal('%')}{::fqdn}"
```

### DIFF
--- a/lib/hiera/recursive_guard.rb
+++ b/lib/hiera/recursive_guard.rb
@@ -9,7 +9,7 @@ class Hiera::RecursiveGuard
   end
 
   def check(value, &block)
-    if @seen.include?(value) and value !~ "literal('\w')"
+    if @seen.include?(value) and value != "literal('%')"
       raise Hiera::InterpolationLoop, "Lookup recursion detected in [#{@seen.join(', ')}]"
     end
     @seen.push(value)
--- a/lib/hiera/recursive_guard.rb
+++ b/lib/hiera/recursive_guard.rb
@@ -9,7 +9,7 @@ class Hiera::RecursiveGuard
   end
 
   def check(value, &block)
-    if @seen.include?(value)
+    if @seen.include?(value) and value !~ "literal('\w')"
       raise Hiera::InterpolationLoop, "Lookup recursion detected in [#{@seen.join(', ')}]"
     end
     @seen.push(value)


kylog:
Minor hygiene things: can you submit a ticket at https://tickets.puppetlabs.com/browse/HI and reference that ticket in the commit summary like so: `(HI-123) Allow yada yada yada ...`.

Hygiene aside, I'll defer to @hlindberg or @thallgren on the change itself.

thallgren:
The problem is that the value is added to the stack more than once in the same interpolation resolution. It shouldn't be, since that's not a recursion per se. This PR doesn't fix that problem (it only deals with '%').

cwood:
yea I know I assumed this wasnt right. I just wanted to get it somewhere so I can make you aware of this issue.

hlindberg:
We need to treat any change to hiera interpolation with great care since it is quite brittle and any change may be disruptive and cause breakage. It is a known problem that multiple literals does not work, (along with numerous other issues related to interpolation and interpolation functions). The long term direction is to use the new implementation of hiera data support in "lookup". Part of the brittleness in Hiera is that there is no comprehensive test suite and it is therefore easy to introduce regressions that goes without detection until it blows up for a number of users.

That is a long way of saying that we are not very likely to alter the behaviour of Hiera (the separate project) as it will be replaced by "lookup". For most problems (like this one), the community has found ways around them (discussed in various tickets). Suggest looking at existing tickets and/or ask for help on the Slack or IRC channels how to work around the problems.

Since this PR also does not address the fundamental problem with hiera interpolation and the literal interpolation function, I am going to close this.

cwood:
@hlindberg this seems sort of rude and long way of saying wont fix. This is a bug thats affecting me and obviously others. If you are not going to fix this it would be good to have some sort of doc or another way to do what I want to do.

hlindberg:
@cwood I am sorry you feel as if I were being rude - that was not my intention. There is unfortunately no "works in every version" work around to the problem - that is why I suggested hopping onto Slack and asking for help there (a dialog is required, figuring out what you need to do, which version you are on, etc.). I am sorry this is not working as expected - in the hiera project unfortunately many aspirations to fix problems and make it better actually have made things worse as holes in the implementation have been exploited as work arounds and become the "defacto" way of solving certain needs. I wish it was otherwise, and unfortunately I am usually the person that ends up having to deliver this unhappy message.

PR #364, opened 2016-09-28
Adding a dynamic yaml backend

brianhks:
The reason for this backend is so that an external system can dynamically create inherited environments.

In the configuration you specify a property as being dynamic using the dynamic_prop field.  The code then can handle the property as being a list.

So for example dynamic_prop is set to env and my enc returns env set to [ qa, lab ]

The backend will dynamically extend the lookup to first got to qa and then to lab.


er0ck:
ping @hlindberg

hlindberg:
Thanks for the interesting idea. There are a couple of things that makes this difficult to accept.
* Hiera is coming up to its end - we are no longer adding new feature to "hiera the project" and we are instead focusing on "lookup" which will be moved out of experimental state in Puppet 4.8.0 (according to plan).
* In the version of lookup being released next there is much more flexibility - what you want may be doable quite easily
* Not sure I understand the proposed functionality, it would need to be elaborated on (in a ticket)
* A Jira ticket is required to make the proposal visible and come to a conclusion what is the best approach for the desired functionality
* Then, the commits needs to be labeled with the ticket nummer to be acceptable
* Don't want merge commits in PRs
* The code in this PR fails the test.

I am therefore closing this PR. Happy to continue the discussion on a Jira ticket.

hlindberg:
If you want to use this dynamic yaml backend with hiera and share it with others, you can always put it in a module and make it available on the forge. Things like this does not have to go into puppet (or hiera).

PR #368, opened 2016-10-14
Revert "Revert "(HI-538) Update beaker version to 3.1.0""

McdonaldSeanp:
Reverts puppetlabs/hiera#366

Jenkins pipelines now have ruby 2.3.1


Magisus:
This works with a locally installed beaker 3.1.0 gem for me, so let's put it into production! 👍

PR #370, opened 2016-10-24
(packaging) Bump version to 3.2.2

Magisus:
This updates Hiera's version to 3.2.2 in preparation for the
puppet-agent 1.8.0 release. We had originally updated the version from
3.2.1 to 3.3.0, but the small amount of changes made a Y release
unnecessary.

### DIFF
--- a/lib/hiera/version.rb
+++ b/lib/hiera/version.rb
@@ -7,7 +7,7 @@
 
 
 class Hiera
-  VERSION = "3.3.0"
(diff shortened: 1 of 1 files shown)


MosesMendoza:
@Magisus given this is actually doing the unusual thing of setting the version lower, maybe some additional context in the commit message about why would be good?

Magisus:
Updated

MosesMendoza:
👍

PR #373, opened 2016-11-02
(PA-652) Bump Hiera version to 3.2.3

Magisus:
This bumps the Hiera version to 3.2.3 following the release of 3.2.2 as
part of puppet-agent 1.8.0.

### DIFF
--- a/lib/hiera/version.rb
+++ b/lib/hiera/version.rb
@@ -7,7 +7,7 @@
 
 
 class Hiera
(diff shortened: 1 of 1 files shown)


Magisus:
The commit checker in Travis requires a HI-XXX ticket for this project (as opposed to any public JIRA ticket). Do we want to allow this to fail, or switch to maint? As a follow up, might be nice to update the filter.

PR #359, opened 2016-08-10
(HI-519) Read files in with unicode

MosesMendoza:
Unless supplied an encoding, File.read will read in files with whatever Ruby's
default external encoding is, which on Windows may be an encoding such as
IBM437. Prior to this commit, hiera's FileCache#read_file method would call
File.read without an encoding, which could be problematic when reading files
containing unicode or other characters undefined on such a platform. This
commit updates the #read_file method to explicitly read files in with
BOM|UTF-8.

PR #377, opened 2017-01-12
(HI-519) Read files in with unicode

MosesMendoza:
Unless supplied an encoding, File.read will read in files with whatever Ruby's
default external encoding is, which on Windows may be an encoding such as
IBM437. Prior to this commit, hiera's FileCache#read_file method would call
File.read without an encoding, which could be problematic when reading files
containing unicode or other characters undefined on such a platform. This
commit updates the #read_file method to explicitly read files in with
BOM|UTF-8.

### DIFF
--- a/spec/unit/filecache_spec.rb
+++ b/spec/unit/filecache_spec.rb
@@ -84,6 +84,36 @@ class Hiera
         end
       end
 
+      it "sets the encoding to UTF-8 when reading a file" do
+        begin
+          original_encoding = Encoding.default_external
+          Encoding.default_external = Encoding::ISO_8859_1
+
+          Dir.mktmpdir do |dir|
+            file = File.join(dir, "testing")
+            write_file(file, "my data")
+            expect(@cache.read_file(file).encoding).to eq(Encoding::UTF_8)
+          end
+        ensure
+          Encoding.default_external = original_encoding
+        end
+      end
+
+      it "reads a file with unicode characters" do
+        begin
+          original_encoding = Encoding.default_external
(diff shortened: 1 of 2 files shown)


MosesMendoza:
supercedes #359

hlindberg:
This is targeting stable, while the ticket says HI 3.3.0 which would be master

MosesMendoza:
Thanks @hlindberg - retargeted and rebased

spec/unit/filecache_spec.rb
```diff
@@ -84,6 +84,36 @@ def write_file(file, contents)
...
+      it "reads a file with unicode characters" do
+        begin
+          original_encoding = Encoding.default_external
+          Encoding.default_external = Encoding::ISO_8859_1
+
+          Dir.mktmpdir do |dir|
+            file = File.join(dir, "testing")
+            write_file(file, "\u2603")
+            expect(@cache.read_file(file)).to eq("\u2603")
+          end
```
Iristyle: As a point of note, in the existing code ... the file bytes are actually read correctly, but they are not considered the appropriate encoding:


```
[1] pry(#<RSpec::ExampleGroups::HieraFilecache::ReadFile>)> foo = @cache.read_file(file)
=> "â\x98\x83"
[2] pry(#<RSpec::ExampleGroups::HieraFilecache::ReadFile>)> foo.bytes
=> [226, 152, 131]
[3] pry(#<RSpec::ExampleGroups::HieraFilecache::ReadFile>)> bar = foo.dup.force_encoding(Encoding::UTF_8)
=> "\u2603"
```

PR #390, opened 2017-03-02
(HI-519) Read files in with unicode

joshcooper:
Unless supplied an encoding, File.read will read in files with whatever Ruby's
default external encoding is, which on Windows may be an encoding such as
IBM437. Prior to this commit, hiera's FileCache#read_file method would call
File.read without an encoding, which could be problematic when reading files
containing unicode or other characters undefined on such a platform. This
commit updates the #read_file method to explicitly read files in with
BOM|UTF-8.

Cherrypick of 45bdf75156c8f14104f7a05d0e46a8c7d9d7f4be originally submitted in PR #377

### DIFF
--- a/lib/hiera/filecache.rb
+++ b/lib/hiera/filecache.rb
@@ -49,7 +49,7 @@ class Hiera
     # in processing will be propagated to the caller
     def read_file(path, expected_type = Object)
       if stale?(path)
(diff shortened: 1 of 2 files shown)


hlindberg:
+1 merging

PR #396, opened 2017-05-24
(HI-572) Update hiera.yaml to Version 5 syntax

jcoconnor:
puppet-agent ships with hiera 3 version hiera.yaml causing deprecation
warnings - should ship with version 5

### DIFF
--- a/ext/hiera.yaml
+++ b/ext/hiera.yaml
@@ -1,13 +1,11 @@
 ---
-:backends:
-  - yaml
-:hierarchy:
-  - "nodes/%{::trusted.certname}"
-  - common
-
-:yaml:
-# datadir is empty here, so hiera uses its defaults:
-# - /etc/puppetlabs/code/environments/%{environment}/hieradata on *nix
-# - %CommonAppData%\PuppetLabs\code\environments\%{environment}\hieradata on Windows
-# When specifying a datadir, make sure the directory exists.
-  :datadir:
+version: 5
+defaults:
+  datadir: "/etc/puppetlabs/code/environments/%{environment}/hieradata"
+  data_hash: yaml_data
+hierarchy:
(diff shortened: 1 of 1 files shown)


branan:
A version 5 hiera.yaml isn't meaningful in this repo - hiera 5 is entirely built in to puppet. This is still hiera 3. I think it would make more sense to puppet this in the puppet repo, or just fix up puppet-agent to contain its own default configuration file instead of sourcing it from one of the components

McdonaldSeanp:
I think the latter choice that @branan mentioned makes the most sense. Agent packaging can easily add the default hiera.yaml file (and have it be consistent with hiera 5) and lay it down as part of the agent install specifically. 

TBH this change seems like the exact reason we have the agent.

branan:
@McdonaldSeanp The big reason to put it in puppet, I think, is that we do like to have the ext dirs as places that example files and such can live - it would be kind of nice to have a base v5 hiera.yaml live with the puppet code. This could also possibly allow us to lay it down with puppet's install.rb, which would make source-based installs (and possibly distro packages?) saner.

McdonaldSeanp:
@branan that makes sense. I would request that installation logs or something of that nature write to the host that it's laying down hiera.yaml, so that we can follow the trail in an easy way as to where hiera.yaml in the agent comes from.

Iristyle:
Given all the Hiera 5 code lives in Puppet, I agree that Puppet seems like the place.

Is packaging already sourcing a copy from Hiera during builds?

jcoconnor:
@Iristyle Yes - it pulls it in from the ```hiera``` repo.
Does puppet-5 use the ```hiera``` repo at all ?

jcoconnor:
Closing this issue - will resurrect a new PR in ```puppet``` and ```puppet-agent```

PR #402, opened 2017-08-24
[HI-578] Add check for key as a symbol

dandunckelman:
See https://tickets.puppetlabs.com/browse/HI-578


hlindberg:
I am closing this because the solution here is not to allow symbols as keys but instead make sure that the input is correct. See longer explanation on the ticket.

PR #405, opened 2017-09-16
(maint) Reference actual config version in error message

haus:
Now that hiera 5 configs exist this error message can be thrown from a
hiera 5 config in which case the error message is very confusing. This
commit includes the actual hiera config version in the error message
instead of assuming it will be 4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,12 @@
 language: ruby
 sudo: false
+# Travis Ruby 1.9.3 stack is defaulting to version 1.7.6 of bundler
+# which is incompatible with the Rubygems 2.6.13 security release.
+#
+# TODO: Remove when Travis updates its default:
+#   https://github.com/travis-ci/travis-ci/issues/8357
+before_install:
+  - gem install bundler --version 1.15.4
 bundler_args: --without development
 script: "bundle exec rake $CHECK"
 notifications:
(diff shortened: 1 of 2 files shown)


haus:
Looks like travis is failing with the same issue that puppet just worked around recently. I'll try to pull that fix over.

branan:
This looks like a sane cleanup to me. :+1:

MikaelSmith:
Related to [HI-577](https://tickets.puppetlabs.com/browse/HI-577), but not a fix for it.

adrienthebo:
:+1:

hlindberg:
But the message is not quite correct. You can use hiera 5 for the global hiera.yaml with puppet (since hiera 5 is in puppet). Neither version 4 nor 5 works with the hiera gem.

MikaelSmith:
Will this code be used to load global hiera.yaml from hiera 5 in puppet?

Sounds like we might want different messages for hiera.yaml version 4 and hiera.yaml version 5.

hlindberg:
- henrik

> 18 sep. 2017 kl. 23:58 skrev Michael Smith <notifications@github.com>:
> 
> Will this code be used to load global hiera.yaml from hiera 5 in puppet?
> 
No. It is only used when using the hiera command line tools or using the gem from Ruby. 
> Sounds like we might want different messages for hiera.yaml version 4 and hiera.yaml version 5.
> 
Not really meaningful. Hiera the gem only works with version 3 hiera.yaml.

- henrik

PR #408, opened 2017-10-09
(HI-585) Enable testing system openssl linked PA in CI pipelines on rhel7

jsane:
VM images for RHEL7 use an old patch version of openssl library compared to that
on Centos7 which is used for building PA for rhel platforms. That causes failures
when running hiera acceptance tests on rhel7. They need to be upgraded with
newer version of openssl libraries.

Related to PA-1580 (part of epic PA-1608). Has been tested in PA CI - logs would be attached to PR for PA-1264 (enables linking PA against system openssl).
This does not impact regular agent build using vendored openssl which disregard system
openssl.


jsane:
Closing this PR; it is not needed per comments in the ticket.

PR #410, opened 2018-03-30
(HI-596) Move hiera to running test from beaker-puppet

melissa:
This includes removing all the set up logic from hiera. This is now
stored in beaker-puppet. There is one test that was previously being
executed during the test phase, but is more appropriate to run as a
pre-suite. This ensures it is run before any of the other tests, so that
order isn't as important.


melissa:
jenkins please test this

melissa:
jenkins please test this

joshcooper:
jenkins please test this on aix61-POWERf,aix71-POWERf,aix72-POWERf,redhat7-POWER,redhat7-AARCH64,sles12-POWER,solaris10-SPARC,solaris11-SPARC,ubuntu1604-POWER,ciscoxr-64,amazon6-64,amazon7-64 with agent 1.10.x

joshcooper:
I verified this works on all vmpooler nodes also

PR #413, opened 2018-08-02
(packaging) Use packaging as a gem

mwaggett:
also, unpin rake dependency.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ group :development do
 end
 
 group :development, :test do
-  gem 'rake', "~> 10.1.0"
+  gem 'rake'
   gem 'rspec', "~> 3.3", :require => false
   gem "rspec-legacy_formatters", "~> 1.0", :require => false
   gem 'mocha', "~> 0.10.5", :require => false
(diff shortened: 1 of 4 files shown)


mwaggett:
travis error is because we dropped support for ruby 1.9.3. if we're actually still supporting that, I can put this on an earlier version of packaging.

and the other failure is the Windows thing :/

.travis.yml
```diff
@@ -23,5 +22,3 @@ matrix:
       env: "CHECK=commits"
     - rvm: 2.0.0
       env: "CHECK=commits"
-    - rvm: 1.9.3
-      env: "CHECK=commits"
```
mwaggett: @joshcooper Do I need to put these back too?

Gemfile
```diff
@@ -1,14 +1,28 @@
...
+  gem 'packaging', *location_for(ENV['PACKAGING_LOCATION'] || '~> 0.99')
+end
+
 group :development do
   gem 'watchr'
 end
 
 group :development, :test do
-  gem 'rake', "~> 10.1.0"
+  gem 'rake'
```
joshcooper: same comment about not making this unbounded

PR #420, opened 2018-10-29
(PA-2183) Update acceptance tests to work with Beaker 4

ekinanp:


speedofdark:
The travis fail is because of commit message checking.

PR #424, opened 2018-11-29
Explicitly require beaker hypervisor gems in acceptance Gemfile

caseywilliams:
These are still development dependencies of beaker 4, but they should be required explicitly


caseywilliams:
Forgot that I'd already made a PR for this - closing.

PR #425, opened 2019-06-24
Allow looking up non-hash key with --hash option

trankchung:
Currently doing `hiera LOOKUP_KEY --hash` an Exception is thrown. This is not desired as we'd like the use `--hash` to have deep merge working but when doing so with non-hash LOOKUP_KEY an exception is thrown. This PR is resolves that issue and allow `--hash` option to work with non-hash LOOKUP_KEY keys.

### DIFF
--- a/bin/hiera
+++ b/bin/hiera
@@ -225,6 +225,18 @@ unless options[:verbose]
   Hiera.logger = "noop"
 end
 
-ans = hiera.lookup(options[:key], options[:default], options[:scope], nil, options[:resolution_type])
+# Currently if --hash is used but the lookup key is not a hash an Exception is thrown.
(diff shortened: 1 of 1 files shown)


trankchung:
@puppetcla I get a 502 Bad Gateway when trying to do CLA. Could you please check?

trankchung:
@puppetcla Could you please check the site https://cla.puppet.com? It's returning 502 bad gateway. Thanks.

joshcooper:
Hi @trankchung. Thank you for submitting this PR, however, this repo is deprecated, and is no longer receiving further improvements. The replacement workflow is to use `puppet lookup`, which is found in the https://github.com/puppetlabs/puppet repo. Please give that a try and if that doesn't work, please submit a PR there.

Also sorry about the cla site. FWIW it appears to be ok now.

PR #429, opened 2021-02-13
gemspec: add SPDX compatible license metadata

priv-kweihmann:
so it can be used by compliance tools properly

Signed-off-by: Konrad Weihmann <kweihmann@outlook.com>

### DIFF
--- a/.gemspec
+++ b/.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = Gem::Requirement.new(">= 0") if s.respond_to? :required_rubygems_version=
   s.authors = ["Puppet Labs"]
   s.date = "2013-04-03"
-  s.description = "A pluggable data store for hierarcical data"
+  s.description = "A pluggable data store for hierarchical data"
   s.email = "info@puppetlabs.com"
   s.executables = ["hiera"]
   s.files = ["bin/hiera"]
--- a/.gemspec
+++ b/.gemspec
@@ -22,4 +22,5 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
   s.rubygems_version = "1.8.25"
   s.summary = "Light weight hierarchical data store"
+  s.license = "Apache-2.0"
 end


justinstoller (approved):

GabrielNagy:
I'm not sure whether this version of hiera is used anymore, but we can merge it...

@priv-kweihmann since you're in gemspec territory, can you also fix the spelling of _hierarchical_ on line 17?

priv-kweihmann:
> I'm not sure whether this version of hiera is used anymore, but we can merge it...
> 
> @priv-kweihmann since you're in gemspec territory, can you also fix the spelling of _hierarchical_ on line 17?

Will do, np

GabrielNagy (approved):

ekohl:
Somehow this isn't showing on https://rubygems.org/gems/hiera so I wonder where it fails.

joshcooper:
We have some bespoke packaging automation that uses `ext/project_data.yaml` when building the gem for release. Since hiera doesn't specify a license in https://github.com/puppetlabs/hiera/blob/main/ext/project_data.yaml the resulting gem doesn't have one. We have plans on cleaning this up in PA-5400:

> The puppet, facter and hiera gems have multiple sources of truth for gem metadata. There's .gemspec and ext/project_data.yaml. The former is used when running bundler or resolving gem dependencies using "latest" like export PUPPET_GEM_VERSION=[https://github.com/puppetlabs/puppet]. We have been bitten repeatedly by having multiple sources of truth. See PUP-11844, PA-1077, PUP-7425, PUP-8685
>
> Distros also repackage our gems and expect that the .gemspec is the source of truth.
> 
> The goal is we should be able to use the normal "gem build" workflow to build a gem, including platform specific gems, e.g. x64-mingw32.
>
> It appears we do use “gem build” for facter when releasing tagged versions (see https://github.com/puppetlabs/ci-job-configs/blob/bc56fdefa85e547a5672c4b2c2578d8298d9b5bd/resources/scripts/vanagon-component-artifact-gen.sh#L107C1-L111) however, we may still rely on build_defaults.yaml for nightly facter builds (see https://github.com/puppetlabs/facter/commit/b0adf6fb86cb0148ab9ce411651ed16c2452b480).

PR #432, opened 2022-01-28
(PA-4192) Move from travis to github actions

AriaXLi:


.github/workflows/rspec_tests.yaml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ${{ matrix.cfg.os }}
+    steps:
+      - name: Checkout current PR
+        uses: actions/checkout@v2
+
+      - name: Install ruby version ${{ matrix.cfg.ruby }}
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.cfg.ruby }}
+
```
joshcooper, line 30: I'm guessing we need to `bundle install` after installing ruby and before running `bundle exec rake spec`? Puppet's version of this does: https://github.com/puppetlabs/puppet/blob/4772afa194402a3876069785a178611797a8eb7d/.github/workflows/checks.yaml#L32-L36

where the `gem update ... ` is necessary on Ruby 2.5 but we do it for all rubies.

.github/workflows/rspec_tests.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Checkout current PR
+        uses: actions/checkout@v2
+
+      - name: Install ruby version ${{ matrix.cfg.ruby }}
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.cfg.ruby }}
+      
+      - run: "bundle exec rake spec"
+      - run: "bundle exec rake commits"
```
joshcooper, line 32: I think this is going to run the `rake commits` tasks for each ruby/os combination. One option is to split the check into a separate file like we do in facter and puppet:

https://github.com/puppetlabs/puppet/blob/main/.github/workflows/checks.yaml

.github/workflows/rspec_tests.yaml
```diff
@@ -0,0 +1,32 @@
...
+        cfg:
+          - {os: ubuntu-18.04, ruby: '2.5'}
+          - {os: ubuntu-18.04, ruby: '2.7'}
+          - {os: windows-2019, ruby: '2.5'}
+          - {os: windows-2019, ruby: '2.7'}
+
+    runs-on: ${{ matrix.cfg.os }}
+    steps:
+      - name: Checkout current PR
+        uses: actions/checkout@v2
```
joshcooper, line 24: some of the puppet & facter versions of these use `fetch-depth: 0` like https://github.com/puppetlabs/puppet/blob/main/.github/workflows/checks.yaml#L25 I assume it limits the amount of git history we fetch, so maybe a good thing to add?

.github/workflows/checks.yaml
```diff
@@ -0,0 +1,33 @@
...
+      matrix:
+        cfg:
+          - {check: rake commits, os: ubuntu-18.04, ruby: '2.5'}
+
+    runs-on: ${{ matrix.cfg.os }}
+    steps:
+      - name: Checkout current PR
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
joshcooper, line 20: Sorry I think I lead you astray with my earlier comment. Fetch depth 0 means [fetch all history for all tags and branches](https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches)! I think we can drop it entirely.

AriaXLi: Oooh okay got it! I'll remove those now then

.github/workflows/rspec_tests.yaml
```diff
@@ -0,0 +1,38 @@
...
+          ruby-version: ${{ matrix.cfg.ruby }}
+
+      - name: Update rubygems and install gems
+        run: |
+          gem update --system --silent --no-document
+          bundle config set without packaging development
+          bundle install --jobs 4 --retry 3
+      
+      - run: "bundle exec rake spec"
+      - run: "bundle exec rake commits"
```
joshcooper, line 38: Sorry just saw this, I think we can drop the `rake commits`, since it's in `checks.yaml`

AriaXLi: Oh heck, I was going to remove that and just forgot, I'll change that now!

.github/workflows/checks.yaml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          ruby-version: ${{ matrix.cfg.ruby }}
+
+      - name: Update rubygems and install gems
+        run: |
+          gem update --system --silent --no-document
+          bundle config set without packaging development
+          bundle install --jobs 4 --retry 3
+      - name: Run ${{ matrix.cfg.check }} check
+        run: bundle exec rake commits ${{ matrix.cfg.check }}
```
joshcooper, line 31: sorry last thing, I think we'll end up with `bundle exec rake commits rake commits` after interpolation

joshcooper (approved):

PR #433, opened 2022-02-03
(PA-4191) Add Snyk monitoring

cthorn42:


.github/workflows/snyk_monitor.yaml
```diff
@@ -0,0 +1,29 @@
+---
+name: Snyk Monitor
+on:
+  schedule:
+    - cron: '0 5 * * 1-5'
```
joshcooper, line 5: want to drop this part?

joshcooper:
@cthorn42 one way to test the snyk monitor before merging is to push a feature branch to the puppetlabs/upstream repo instead of your fork. In doing so, your PR will have access to tokens (which still need to be added to this repo). That way we can validate that the snyk monitor check passes end-to-end before merging this PR.

cthorn42:
Okay I need to do a squash, but I got this to work. Only trick was the name of the token/key was different. After updating that I was able to get the action to work, and on Snyk it shows that there is a report under puppet-foss there is a Hiera report.

joshcooper:
Looks like the commit checker is failing. Maybe just prefix the commit summary with `(maint)`

joshcooper (approved):

PR #434, opened 2022-02-07
(PA-4192) Fix packaging error from checks.yaml

AriaXLi:


joshcooper:
ah ok, I see what's going on. By default, the [checkout v2 action](https://github.com/actions/checkout#checkout-v2) doesn't preserve the commit history, so trying to reference `HEAD` won't work. We'll need to add `fetch-depth` like we do in the puppet checks: https://github.com/puppetlabs/puppet/blob/4772afa194402a3876069785a178611797a8eb7d/.github/workflows/checks.yaml#L24-L25

But we only want to do that for the rake commits check, not the rspec checks, since in the latter case we don't need all of the history.

.github/workflows/checks.yaml
```diff
@@ -29,3 +29,5 @@ jobs:
           bundle install --jobs 4 --retry 3
       - name: Run ${{ matrix.cfg.check }} check
         run: bundle exec ${{ matrix.cfg.check }}
+        with:
+          fetch-depth: 0
```
joshcooper, line 33: Ah I think this needs to go with the checkout action

joshcooper (approved):

PR #438, opened 2022-02-21
Set minimal ruby version to 2.5 in gemspec

bastelfreak:
Ruby 2.5 is the oldest version that's tested in CI. the gemspec should
reflect that.

### DIFF
--- a/.gemspec
+++ b/.gemspec
@@ -23,4 +23,5 @@ Gem::Specification.new do |s|
   s.rubygems_version = "1.8.25"
   s.summary = "Light weight hierarchical data store"
   s.license = "Apache-2.0"
+  s.required_ruby_version = '>= 2.5.0'
 end


.gemspec
```diff
@@ -23,4 +23,5 @@ Gem::Specification.new do |s|
   s.rubygems_version = "1.8.25"
   s.summary = "Light weight hierarchical data store"
   s.license = "Apache-2.0"
+  s.required_ruby_version = '>= 2.5.0'
```
joshcooper, line 26: The change makes sense to me, but we've generally only dropped ruby versions in majors (in puppet and bolt) even when the dropped versions are EOL... However, recent versions of bundler take required ruby versions into account when resolving dependencies so it's not as significant an issue as it was before...

bastelfreak: but what's the current supported Ruby version? Ci only runs on Ruby 2.5 and newer so one could argue that this PR doesn't drop support for any ruby versions, it just writes down what's currently supported.

joshcooper: Well just because we don't run test it in CI doesn't mean you can't install and run it on older ruby versions. That said this repo hasn't had meaningful code change since 2017 (just version bumps and Gemfile updates since then), so I'm fine adding this.

joshcooper:
Not sure what's up with the ruby 3 check, I'm going to close and reopen

joshcooper (approved):

PR #444, opened 2023-01-12
(maint) Use keyword arguments

joshcooper:
FileUtils in ruby 3.x doesn't accept passing args as a hash.


mhashizume:
Updated the branch protection rules from Ruby 3.2.0-preview to final for this PR.

mhashizume (approved):

